PR #9, opened 2022-07-13
Pause anvil logs

noahlitvin:
Trying to pass a 'pointer' into the rpc file (`showLogs.rpc`), thought an object would do it? Idk typescript. Otherwise should be all set.


dbeal-eth (changes requested):
why not just return the `child_process` instance from the function and then hook into `stdout` directly from the `index.ts` file? full control and less pointer complexity

PR #12, opened 2022-07-13
Better anvil check

noahlitvin:


packages/cli/src/rpc.ts
```diff
@@ -21,7 +22,7 @@ export function runRpc({ port, forkUrl }: RpcOptions): Promise<ChildProcess> {
 
   return Promise.race<Promise<ChildProcess>>([
     new Promise<ChildProcess>((resolve, reject) => {
-      const anvilInstance = spawn('anvil', opts);
+      const anvilInstance = spawn(os.homedir() + '/.foundry/bin/anvil', opts);
```
dbeal-eth, line 25: is this change still needed?

PR #13, opened 2022-07-14
Guides, etc.

noahlitvin:
- Main addition is the `--registry-address` option. Let's do a release after this is in.
- We're defaulting to the cloudflare rpc in the cli, so I switched it to this in the hardhat plug-in
- Wrote the deployments guide
- Removed `publisherPrivateKey`, doesn't seem to be used.
- Made the `anvil: ` part of the CLI output gray. It's still rendering twice sometimes, but didn't nail down why yet.


dbeal-eth (approved):
isnt the `/docs` page getting a bit too long? maybe just push these guides mostly to a blog and we can link them as we go

PR #14, opened 2022-07-14
[WIP] Contracts project

noahlitvin:
* Adds chainlink mock from snx v2 codebase
  * Note that I added a function we may not want to keep
  * @dbeal-eth mentioned we probably don't want reverts in the views
  * Here's a mock from a chainlink repo: https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/tests/MockV3Aggregator.sol ("We currently have the v0.6 versions deployed so best to use those for testing." https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.6/tests/MockV3Aggregator.sol)
* Removes duplicate `import 'hardhat-interact';`
* Hitting an issue with getting the `HardhatUserConfig` type to accept the `cannon` key?

### DIFF
--- a/packages/contracts/cannonfile.chainlink.toml
+++ b/packages/contracts/cannonfile.chainlink.toml
@@ -1,0 +1,29 @@
+name = "chainlink"
+version= "0.1.0"
+description = "Official Cannon build for Chainlink mock aggregegator"
+keywords = ["chainlink"]
+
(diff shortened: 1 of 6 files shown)


noahlitvin:
Will revisit with V2

PR #17, opened 2022-07-16
Fix Onboarding

noahlitvin:
- [x] Hitting an issue with `lerna link`, seems like I can't get the cli to build since it's not picking up the new helpers package.
- [ ] Test CLI after downgrading foundry
- [ ] Test hardhat plug-in after downgrading foundry

Couldn't identify the best command to load an older version to test. Should be something like this: https://github.com/foundry-rs/foundry/tree/master/foundryup (We need any release that has an earlier timestamp than `anvil 0.1.0 (37e4376 2022-07-16T00:05:51.565487Z)`)


noahlitvin:
Tried downgrading with `foundryup -C 94bfdb2` to test but got `error: `/Users/noahlitvin/.foundry/foundry-rs/foundry/anvil` is not a directory. --path must point to a directory containing a Cargo.toml file.
foundryup: command failed: cargo install --path ./anvil --bin anvil --locked --force --root /Users/noahlitvin/.foundry`

Other than testing (and fixing the CI test?) this should be all set.

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    if (cargoResponse.confirmation) {
+      console.log(magentaBright('Installing Cargo...'));
+      await execPromise('curl https://sh.rustup.rs -sSf | sh');
+    } else {
+      process.exit();
+    }
+  }
+
+  const MINIMUM_NIGHTLY_ANVIL_BUILD_TIMESTAMP = 1657679573421;
```
dbeal-eth, line 24: yay, though you can probably truncate that down to a second and add a comment clarifying what date we are referring to here

packages/cli/src/helpers.ts
```diff
@@ -0,0 +1,98 @@
...
+import prompts from 'prompts';
+import { magentaBright } from 'chalk';
+
+export async function setupAnvil(): Promise<void> {
+  const needsCargo = !(await hasCargo());
+  if (needsCargo) {
+    const cargoResponse = await prompts({
+      type: 'confirm',
+      name: 'confirmation',
+      message: 'Cannon requires Cargo (https://doc.rust-lang.org/cargo/). Install it now?',
```
dbeal-eth, line 12: cargo is probably actually not needed now that we have been upstreamed (cargo is a build tool used by rust, but you dont need to build anything). At least, its not needed by the synthetix cli anymore

dbeal-eth (changes requested):
please review ci as well, the errors are certainly unusual now

PR #18, opened 2022-07-17
Minor UX Improvements

noahlitvin:
Improves CLI + Hardhat plug-in output.

### DIFF
--- a/packages/hardhat-cannon/src/tasks/publish.ts
+++ b/packages/hardhat-cannon/src/tasks/publish.ts
@@ -1,5 +1,6 @@
 import path from 'path';
 import { task } from 'hardhat/config';
+import { bold, yellowBright } from 'chalk';
 
 import { CannonRegistry } from '@usecannon/builder';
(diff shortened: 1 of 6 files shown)


packages/hardhat-cannon/src/tasks/publish.ts
```diff
@@ -27,7 +33,7 @@ task(TASK_PUBLISH, 'Provision and publish to the registry the current Cannonfile
     const response = await prompts({
       type: 'confirm',
       name: 'confirmation',
-      message: `This will deploy your package to IPFS and use ${signers[0].address} to add the package to the registry. Continue?`,
+      message: `This will deploy your package to IPFS and use ${signers[0].address} to add the package to the registry. (This will cost a small amount of gas.) Continue?`,
```
dbeal-eth, line 36: would prefer if we add the functionality to add the exact gas estimate and an amount of eta estimated to be spent. merge this now and add it as an issue?

dbeal-eth (approved):

Issue #7, opened 2022-07-12
Be able to import a project directly from an IPFS/http url (not use registry)


noahlitvin:
originally https://github.com/usecannon/cannon-legacy/issues/6

dbeal-eth:
also add capability to upload packages without actually registering them with registry

dbeal-eth:
done in the 1.1 relaeses

PR #42, opened 2022-10-04
Add load from ipfs

dbeal-eth:
allow for resolving packages from a IPFS URL, bypassing the registry. This works by updating the `CannonRegistry` to only optionally take its parameters.
Then, if it encounters a cannon package named `@ipfs` (note `@` symbol
is disallowed on registry), it will
automatically skip resolve from registry and instaed load directly from
the IPFS hash specified in the `version` field. It will also associate a
link for this package so that it can be resolved internally without
issue, as is customary.

To allow for users to leverage this functionality, they can push with
only.

In the future, we could expand this naming scheme to allow for resolving
from `@file` or `@url`.

implements https://github.com/usecannon/cannon/issues/7

note: do not merge (or at least don't release), this probably breaks
hadrhat-cli since it seems to depend on the functionality to load from
IPFS using `fetch`, but that feature breaks in `hardhat-cannon` which
expects fully fledged IPFS API, which has a different API from the
regular fetch.


packages/builder/src/registry.ts
```diff
@@ -86,28 +91,37 @@ export class CannonRegistry {
...
-    const bufs: Uint8Array[] = [];
-
     debug(`downloading content from ${this.url}/${hash}`);
 
-    const response = await fetch(`${this.url}/${hash}`);
-    const blob = await response.blob();
-    const arrayBuffer = await blob.arrayBuffer();
-    return Buffer.from(arrayBuffer);
+    const chunks = [];
+    for await (const chunk of this.ipfs.cat(urlOrHash)) {
```
mjlescano, line 97: I think this is revering a change made by @noahlitvin yesterday

mjlescano (approved):

packages/builder/src/registry.ts
dbeal-eth: yep known issue

dbeal-eth:
@mjlescano could you re-review? if everything looks good this should be safe to merge and release

PR #51, opened 2022-10-24
need ability to use `abiOf` with an array of previous contracts

dbeal-eth:
for "diamond" style proxies which have multiple implementations behind the main, it is required to supply an array for `abiOf` instaed of a single target.

Additionally, (this is extremely minor, but) it should be possible to specify the `from` address on the `contract` step

### DIFF
--- a/packages/builder/src/steps/contract.ts
+++ b/packages/builder/src/steps/contract.ts
@@ -141,8 +141,7 @@ export default {
       }
 
       abi = JSON.parse(implContract.interface.format(ethers.utils.FormatTypes.json) as string);
-    }
-    else if (config.abiOf) {
+    } else if (config.abiOf) {
       abi = [];
       for (const ofContract of config.abiOf) {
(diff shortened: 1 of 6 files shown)


packages/builder/src/steps/contract.ts
```diff
@@ -133,6 +141,17 @@ export default {
...
+    } else if (config.abiOf) {
+      abi = [];
+      for (const ofContract of config.abiOf) {
+        const implContract = getContractFromPath(ctx, ofContract);
+
+        if (!implContract) {
+          throw new Error(`previously deployed contract with identifier "${ofContract}" for factory not found`);
+        }
+
+        abi.push(...JSON.parse(implContract.interface.format(ethers.utils.FormatTypes.json) as string));
```
mjlescano, line 153: should we check for repeated abi fragments here?

mjlescano (approved):

PR #57, opened 2022-10-25
Graceful verification error

noahlitvin:


packages/hardhat-cannon/src/tasks/verify.ts
```diff
@@ -50,18 +51,20 @@ task(TASK_VERIFY, 'Verify a package on Etherscan')
...
-          contract: `${outputs.contracts[c].sourceName}:${outputs.contracts[c].contractName}`,
-          address: outputs.contracts[c].address,
-          constructorArguments: outputs.contracts[c].constructorArgs || [],
-        });
-      } catch (err) {
-        if ((err as Error).message.includes('Already Verified')) {
-          console.log('Already verified');
-        } else {
-          throw err;
+      if (_.get(outputs, ['contracts', c, 'sourceName']) && _.get(outputs, _.get(outputs, ['contracts', c, 'contractName'])) {
```
dbeal-eth, line 54: there should be an `else` block which explains to the user that verification has been skipped for this contract because `sourceName / contractName for ${c} has not been set` or so

packages/hardhat-cannon/src/tasks/verify.ts
```diff
@@ -50,18 +51,20 @@ task(TASK_VERIFY, 'Verify a package on Etherscan')
...
+        console.log('Verifying contract:', c);
+        try {
+          await hre.run('verify:verify', {
+            contract: `${outputs.contracts[c].sourceName}:${outputs.contracts[c].contractName}`,
+            address: outputs.contracts[c].address,
+            constructorArguments: outputs.contracts[c].constructorArgs || [],
+          });
+        } catch (err) {
+          console.log(
+            `Unable to verify ${outputs.contracts[c].sourceName}:${outputs.contracts[c].contractName} - ${
```
dbeal-eth, line 64: ordinarily I would suggest doing the error parsing like we were doing before but as we saw the upstream error could change or be different depending on the release so probably just failing gracefully here for now is indeed the best overall path.

dbeal-eth (changes requested):

packages/hardhat-cannon/src/tasks/verify.ts
dbeal-eth: also rather than wrapping everything in yet another deeper `if` block, why not just add a small `if` block at the beginning that checks if these properties exist if not, then `continue` out

PR #68, opened 2022-10-29
Fix import settings

dbeal-eth:
if a package is `import`ed from another cannonfile, it should default to re-using the same settings used by the previous deployment. This would result in the behavior where, if a user specifies no override `options` on deployment, the package is automatically returned-as-is (aka, with no  upstream invocations on import).


noahlitvin (approved):

packages/builder/src/steps/import.ts
```diff
@@ -40,6 +44,8 @@ export default {
   configInject(ctx: ChainBuilderContext, config: Config) {
     config = _.cloneDeep(config);
 
+    console.log(config, ctx);
```
mjlescano, line 47:
```suggestion
```

PR #74, opened 2022-11-14
fix default overrides on registry publish

mjlescano:
This PR fixes a bug when trying to publish a package without custom settings it wouldn't work, because ethers.js would think that the last `undefined` was an extra argument for the contract call.


dbeal-eth (approved):
did you test/verify that adding an empty object here provides the desired fix on your machine?

PR #85, opened 2022-12-08
fix get contract from path function bug

dbeal-eth:


packages/builder/src/util.ts
```diff
@@ -139,7 +139,7 @@ export function getContractDefinitionFromPath(ctx: ChainBuilderContext, path: st
 
   let importsBase: ChainArtifacts = ctx;
   for (const p of pathPieces.slice(0, -1)) {
-    importsBase = ctx.imports[p];
+    importsBase = importsBase.imports![p];
```
dbeal-eth, line 142: tiny fix lol

noahlitvin (approved):

PR #98, opened 2023-01-25
Updates to the registry

mjlescano:
## Changes
- [x] Remove misc contracts in favor of using directly the `@synthetixio/core-contracts` dependency, this will include the latest updates and remove some boilerplate from our code
- [x] Update tests to include latest changes


dbeal-eth (approved):
thanks for updating this

PR #104, opened 2023-01-30
Finish guides

noahlitvin:


dbeal-eth (approved):
both sections need a testing section added, but other than that looking pretty good

PR #123, opened 2023-02-10
Fix browser build

mjlescano:


packages/builder/src/registry.ts
```diff
@@ -35,11 +35,14 @@ export class OnChainRegistry extends CannonRegistry {
     overrides = {},
   }: {
     address: string;
-    signerOrProvider: ethers.Signer | ethers.providers.Provider;
+    signerOrProvider: string | ethers.Signer | ethers.providers.Provider;
```
dbeal-eth, line 38: why this change?

packages/builder/src/util.ts
```diff
@@ -1,13 +1,14 @@
 import crypto from 'crypto';
 import { ethers } from 'ethers';
-
+import { Buffer } from 'buffer';
 import _ from 'lodash';
+import { JsonFragment } from '@ethersproject/abi';
```
dbeal-eth, line 5: why this change

dbeal-eth (changes requested):

packages/cli/src/util/loader.ts
```diff
@@ -0,0 +1,9 @@
+import { CannonRegistry, IPFSLoader as CannonIPFSLoader } from '@usecannon/builder';
+
+export class IPFSLoader extends CannonIPFSLoader {
+  constructor(ipfsUrl: string, resolver: CannonRegistry) {
```
dbeal-eth, line 4: kind of a wierd way to do it, would rather just have a function like `getIpfsLoader()`

dbeal-eth (approved):

PR #126, opened 2023-02-16
Add warning when building on hardhat network

mjlescano:


dbeal-eth (approved):
change final part of the warning to say `If this is not what you want, consider using `--network cannon`)

thank you :pray:

PR #138, opened 2023-03-07
invoke target as string (not only array)

noahlitvin:
is this impossible to do with jtd?


noahlitvin:
Closing in favor of #364

PR #142, opened 2023-03-07
abiOf notice

noahlitvin:


packages/builder/src/steps/invoke.ts
```diff
@@ -94,19 +94,26 @@ async function runTxn(
...
+  const abiOfNotice =
+    ' If this is a proxy contract, make sure you’ve specified abiOf for the contract action in the cannonfile that deploys it.';
+
   if (!contract.functions[config.func]) {
     throw new Error(
       `contract ${contract.address} for ${currentLabel} does not contain the function "${
         config.func
-      }". List of recongized functions is:\n${Object.keys(contract.functions).join('\n')}`
+      }". List of recongized functions is:\n${Object.keys(contract.functions).join('\n')}${
+        currentLabel.toLowerCase().includes('proxy') ? abiOfNotice : ''
```
dbeal-eth, line 105: this is reliable for synthetix but many places may come up with other names besides proxy or just not use proxy at all (as we have done in the past). Maybe just display it always?

dbeal-eth (approved):

packages/builder/src/steps/invoke.ts
noahlitvin: Yea that's fair. Also noticing the typo on recongized here

PR #143, opened 2023-03-07
Context in error

noahlitvin:


dbeal-eth (commented):
interesting placement, seems like maybe the fix should be going in the `runStep` function instead? Then again I guess, as long as the context is avaiable, we should be printing it out.

noahlitvin:
Just didn't think of this. It'd be cleaner in `runStep`. Should I move it there? Not really sure what the change would mean practically.

dbeal-eth (approved):

PR #167, opened 2023-04-17
better build ux

noahlitvin:
Screenshot 2023-04-16 at 8 38 24 PM


dbeal-eth (approved):
this has been needing an update for awhile! however, lots of lint issues to fix before this can be merged.

dbeal-eth:
not sure why the `lint` job is marking everything as OK, but if you look there are a ton of lints that should be resolved.

PR #177, opened 2023-04-25
fix wipe

dbeal-eth:
wipe has not been working because the old state is getting passed into the builder without reading the `wipe` option passed in (despite the logging printing out the appropriate output)


noahlitvin (approved):

noahlitvin:
Good catch

PR #185, opened 2023-05-04
Add decode command on CLI

mjlescano:
Adds a decode command for function calls, can be tested out with:

```bash
ts-node bin/cannon.ts decode synthetix 0x11e72a430000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000145370617274616e20436f756e63696c20506f6f6c000000000000000000000000
```

And, it will show a message like:
```bash
setPoolName(uint128,string) 0x11e72a43
  poolId: 1
  name: "Spartan Council Pool"
```

### DIFF
--- a/packages/builder/src/registry.ts
+++ b/packages/builder/src/registry.ts
@@ -280,7 +280,7 @@ export class OnChainRegistry extends CannonRegistry {
 
     if (baseResolved) return baseResolved;
 
-    const [name, version] = packageName.split(':');
+    const [name, version = 'latest'] = packageName.split(':');
 
     const url = await this.contract.getPackageUrl(
       ethers.utils.formatBytes32String(name),
@@ -296,7 +296,7 @@ export class OnChainRegistry extends CannonRegistry {
 
     if (baseResolved) return baseResolved;
 
-    const [name, version] = packageName.split(':');
+    const [name, version = 'latest'] = packageName.split(':');
 
     const url = await this.contract.getPackageMeta(
       ethers.utils.formatBytes32String(name),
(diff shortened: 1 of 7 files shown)


dbeal-eth:
awesome! please add supports for:
* decoding errors and events
* decoding nested contracts

dbeal-eth:
hey! I was running some tests to see how well the command does. Noticing some issues, unfortunately:

* I tried the sample/test command you provided up above, but it doesn't work for me (returns `Error: deployment data could not be downloaded for synthetix at 13370-main`)
* for nested function decode, when I run a command like `node packages/cli/dist/bin/cannon.js decode synthetix-omnibus:latest --chain-id 5 0x633026dd73444f474550455250000000000000000000000000000000000000000000000000000000000000000000000000000000000000000039e7139a8c08fa06000000` (which should decode to `setMaxMarketValue(bytes32 _marketKey,uint256 _maxMarketValue)` on SpotMarketProxy), it says that its not found. When I try to decode ``
* it appears `:latest` is not implied when the package name is specified without a tag
* it should be possilble to decode `Error(string)` and `Panic(uint)` opcodes always, just in case the user inputs them without realizing what they are. you can see a sample of this in the `builder/src/error/index.ts` file, its pretty slick.

noahlitvin:
This ultimately worked, but got some weird console output as well:
Screenshot 2023-05-23 at 9 57 58 PM

noahlitvin (approved):

Issue #213, opened 2023-05-26
interact does not parse events, errors (but works in simulate mode)


dbeal-eth:
when using `cannon interact`, errors and events are not processed.

also, some events are not parsed even in `cannon run` mode

PR #214, opened 2023-05-26
fix parsing of some errors and logs in the interact command

dbeal-eth:
fixes https://github.com/usecannon/cannon/issues/213

the general error here was a combo of:
* logs not being parsed correctly for cases where addresses are duplicated in contract list
* the `signers` that come back from frame do not pass through the error parser in cannon (and there probably isnt a way to fix this), but this is unnecessary so use raw `provider` instead
* frame error code was not in the error parser module

also a few more enhancements added:
* print out calldata and gas usage on read-only txns
* parse decimal amounts of ETH when specifying a payable amount
* ability to input `tuple` data types as arguments to a function

### DIFF
--- a/packages/builder/src/error/index.ts
+++ b/packages/builder/src/error/index.ts
@@ -32,7 +32,7 @@ export async function handleTxnError(
 
   if (err.code === 'UNPREDICTABLE_GAS_LIMIT') {
     return handleTxnError(artifacts, provider, err.error);
-  } else if (err.code === 'CALL_EXCEPTION') {
+  } else if (err.code === 'CALL_EXCEPTION' || err.code === 3) {
     txnData = err.transaction;
     errorCodeHex = err.data;
   } else if (err.code === -32015) {
--- a/packages/cli/package-lock.json
+++ b/packages/cli/package-lock.json
@@ -1,2920 +1,2798 @@
 {
-  "name": "@usecannon/cli",
-  "version": "2.4.14",
-  "lockfileVersion": 2,
-  "requires": true,
-  "packages": {
-    "": {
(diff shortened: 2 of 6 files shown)


aminlatifi:
@dbeal-eth Can you describe here how can I test this? I like be able to generate different custom scenario myself.

dbeal-eth:
> @dbeal-eth Can you describe here how can I test this? I like being able to generate different custom scenarios myself.

yes of course! you can test this PR by using the interact command with any package. A good package to test more of the edge cases with is Synthetix:

```
cannon synthetix -c 10
```

...for example, to test with synthetix on optimism mainnet

you can now run a transaction to test the logs fix. for example, if you pick the `CoreProxy` cotnract, and then pick `createAccount(uint128)`, you should see the events displayed for the account creation and decoded properly


you can test printing of calldata and gas usage by calling any read-only function, such as `owner` in `CoreProxy`

you can test parsing of decimal amounts for payable by calling any payable function (not available on this version of synthetix, I would need to find a package, or you could modify the greeter?)

you can test ability to input tuple data types by calling `configureCollateral`, but you will need to type an entire json data structure, which can be a bit difficult. but at least its possible

aminlatifi:
- [x] run a transaction to test the logs fix
    image
- [x] printing of calldata and gas usage by calling any read-only function
    image
- [x] parsing of decimal amounts for payable by calling any payable function I don't see difference between this branch and the package, but everything looks fine

- [x] ability to input tuple data

aminlatifi (approved):
Looks good. The decoding logic is working well, tried to debug and everything looks fine.

Issue #203, opened 2023-05-23
CLI cannot resolve artifacts at custom path


aminlatifi:
When building a cannonfile with a custom  /out/ folder (with --artifacts-directory or -a switch) from a different root folder, it will try to look for the Library artifacts in ./out instead of the one you give it
![image](https://github.com/usecannon/cannon/assets/5684607/781a1162-2a61-4afc-8728-1c97d1058212)

PR #216, opened 2023-05-27
CLI build non-local cannon file support

aminlatifi:
Made cli build to support cannon file in another path
Removed misleading contracts-directory and artifacts-directory options from cli build command 
Removed unused variables
Ref #203

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -53,7 +53,6 @@ export async function build({
   getArtifact,
   getSigner,
   getDefaultSigner,
-  projectDirectory,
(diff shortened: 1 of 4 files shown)


aminlatifi:
`--contracts-directory` and `--artifacts-directory` are misleading options in the cli build command. These options are not used at all.
Moreover, the cli build didn't non-local cannon files. I implemented this supporting and removed those misleading options.

noahlitvin:
Great! Could you update the docs to remove these params as well?

aminlatifi:
@noahlitvin I did

dbeal-eth (approved):

PR #223, opened 2023-06-01
prevent stalled steps by implementing a basic timeout system

dbeal-eth:
on rare occasion jsonrpc APIs will just send a transaction and then it never advances/gets picked up, but cannon still thinks it was successfully submitted.

when this occurs, it can cause cannon to be stuck indefinitely, and the user eventaully has to control-c to cancel the build. this is very bad, since control-c will lead to a loss of state and could impact future builds.

to prevent this, a global step timeout system is implemented. if a step exceeds its given time limit, its promise is immediately suspended and the step is marked as incomplete with `timeout without error`. the user can just retry after the build is complete if it really is just a dumb timeout issue with the api.

### DIFF
--- a/packages/builder/src/actions.ts
+++ b/packages/builder/src/actions.ts
@@ -34,6 +34,8 @@ export interface CannonAction {
     optionalProperties?: Record<string, SomeJTDSchemaType>;
     additionalProperties?: boolean;
   };
+
+  timeout?: number;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


dbeal-eth:
to test this PR, one can simulate transactions which are not being mined by creating an anvil server in no mining mode, adn then running a cannon build against it. For this example, we use the greeter.

in one terminal,
```
anvil --no-mining --chain-id 1337
```

then in the greeter project,
```
PRIVATE_KEY=0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 npx hardhat cannon:build --network local
```

sample output:
image

mohammadranjbarz (approved):
Thanks @dbeal-eth , LGTM

PR #227, opened 2023-06-05
prevent printing of `object Object` on skipped step

dbeal-eth:
if a step is skipped, it is critical we print out hte actual reason why the step failed.

the change here is simple, if the error is just going to be `[object Object]`, stringify the object instead

result:
image

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -114,7 +114,9 @@ export async function build({
   runtime.on(Events.PreStepExecute, (t, n, _c, d) => console.log(`${'  '.repeat(d)}exec: ${t}.${n}`));
   runtime.on(Events.SkipDeploy, (n, err, d) => {
     partialDeploy = true;
-    console.log(`${'  '.repeat(d)}  -> skip ${n} (${err.toString()})`);
+    console.log(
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -114,7 +114,9 @@ export async function build({
   runtime.on(Events.PreStepExecute, (t, n, _c, d) => console.log(`${'  '.repeat(d)}exec: ${t}.${n}`));
   runtime.on(Events.SkipDeploy, (n, err, d) => {
     partialDeploy = true;
-    console.log(`${'  '.repeat(d)}  -> skip ${n} (${err.toString()})`);
+    console.log(
+      `${'  '.repeat(d)}  -> skip ${n} (${err.toString() === '[object Object]' ? JSON.stringify(err) : err.toString()})`
```
aminlatifi, line 118: @dbeal-eth Can you try this? if it works would support more general types
```suggestion
      `${'  '.repeat(d)}  -> skip ${n} (${typeof err === 'object' ? JSON.stringify(err) : err.toString()})`
```

aminlatifi (changes requested):

packages/cli/src/commands/build.ts
dbeal-eth: im just worried that objects that have a proper `toString` implementation will end up still getting stringified even though they printo ut something resasonable/desired for `toString()`

Issue #184, opened 2023-05-02
`cannon build` should infer `chainId` when `--provider-url` is specified


dbeal-eth:
When running cannon build on a non cannon network, and `--provider-url` is specified without specifying `--chain-id`, `chainId` should be set to the resolved chainId from `getNetwork` call on provider.

aminlatifi:
<img width="551" alt="image" src="https://github.com/usecannon/cannon/assets/5684607/a346c250-8b3d-40d3-bb1e-18f6369ba877">

If the above scenario is the expectation, the logic has already been implemented and there is no need for any change.
@dbeal-eth  It seems `--provider-url` isn't used without `--chain-id`. If it's right, I think we can close this issue.

dbeal-eth:
the idea of this issue is that it is possible to *derive* `--chain-id` from `--provider-url`, if it is supplied.

for example, right now if a user who is not using Frame wants to deploy to goerli, they have to type (for example):

```
cannon build cannonfile.toml --chain-id 5 --provider-url https://goerli.infura.io/arstarstarsta
```

in this case, it is annoying to have to type `--chain-id`, because technically cannon could figure out that you want to deploy to `--chain-id 5` just from the specification of the `--provider-url` argument, since the chain id can be queried from the provider with `provider.getNetwork()` call. so with this change, we would expect to be able to run a command like:

```
cannon build cannonfile.toml --provider-url https://goerli.infura.io/arstarstarsta
```

dbeal-eth:
`--chain-id` is still useful if buildign with frame, or if you are on CI and want to add a sanity check to make sure you are on the correct network

PR #229, opened 2023-06-06
Load chainId from providerUrl on build

aminlatifi:
Ref #184


packages/cli/src/index.ts
```diff
@@ -146,17 +146,22 @@ async function doBuild(cannonfile: string, settings: string[], opts: any): Promi
...
-    chainId = (await provider.getNetwork()).chainId;
   } else {
-    const p = await resolveWriteProvider(cliSettings, opts.chainId);
-    chainId = (await p.provider.getNetwork()).chainId;
+    if (opts.providerUrl && !opts.chainId) {
+      const _provider = new ethers.providers.JsonRpcProvider(opts.providerUrl);
+      chainId = (await _provider.getNetwork()).chainId;
+    }
+    const p = await resolveWriteProvider(cliSettings, chainId || opts.chainId);
+    if (!chainId) {
```
dbeal-eth, line 162: this code here is kind of unusual... why not just add an `else` to the if above and have the statement set `chainId = opts.chainId`? Then you can remove `chainId || opts.chainId` from line 161

dbeal-eth (changes requested):

packages/cli/src/index.ts
aminlatifi: Good point, I though first it's mandatory to resolve the chainId from provider. But you say we can rely on what user has set.

Issue #231, opened 2023-06-08
Clean up error messaging


noahlitvin:
Let's hide the stack trace and just show the error being thrown in the CLI. e.g. just show "Error: deployment not found: cannon-omnibus:latest. please make sure it exists for the variant 13370-main." instead of:
<img width="950" alt="Screenshot 2023-06-08 at 2 16 22 PM" src="https://github.com/usecannon/cannon/assets/335975/de9eca64-3d68-4cf8-9462-cfbafc7a2d08">

PR #235, opened 2023-06-11
Handle asynchronous commands in cli error handling

aminlatifi:
Ref #231

### DIFF
--- a/packages/cli/bin/cannon.ts
+++ b/packages/cli/bin/cannon.ts
@@ -9,7 +9,7 @@ cli
     process.exit(0);
   })
   .catch((err) => {
-    if (err.message) {
+    if (err.message && process.env.TRACE !== 'true') {
       console.error(red('Error: ' + err.message));
     } else {
       console.error(err);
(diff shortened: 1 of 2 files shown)


dbeal-eth:
can we add a way to be able to enable full stack traces by environment variable (ex. `TRACE=true` or whatever?

mohammadranjbarz:
*  With this command you can get error from published package
`cannon inspect cannon-omnibus --write-deployments ./deployments`

*  With below command you can execute it from this PR to see it works ( before running that make sure you have built the packages with this command `npx run build`)
* `ts-node packages/cli/bin/cannon.ts  inspect cannon-omnibus --write-deployments ./deployments`

* In this case you can pass `TRACE=true` as environment variable
` TRACE=true ts-node packages/cli/bin/cannon.ts  inspect cannon-omnibus --write-deployments ./deployments`
Screen Shot 1402-03-24 at 12 51 09


JFYI @aminlatifi  @dbeal-eth

dbeal-eth (approved):

Issue #233, opened 2023-06-09
`cannon decode` fails `invalid bytes32 string - no null terminator` when a function parameter is bytes32 with no null term


dbeal-eth:
looks like decode is trying to decode bytes32 strings into regular strings and failing, causing a failure during decode

```
➜  synthetix-deployments git:(xchain-pools) ✗ cannon decode synthetix-omnibus:latest --chain-id 11155111 --preset xchain 0x0ca76175c5c7f4f7e3707a24d13be2e5574c75a457bc9765cad7b06edd12f68bf2c275a2000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001243424f522070617273696e67206572726f720000000000000000000000000000

function handleOracleFulfillment(bytes32 requestId, bytes response, bytes err) 0x0ca76175
/usr/lib/node_modules/@usecannon/cli/node_modules/@ethersproject/strings/lib/bytes32.js:25
        throw new Error("invalid bytes32 string - no null terminator");
              ^
Error: invalid bytes32 string - no null terminator
    at Object.parseBytes32String (/usr/lib/node_modules/@usecannon/cli/node_modules/@ethersproject/strings/lib/bytes32.js:25:15)
    at _renderValue (/usr/lib/node_modules/@usecannon/cli/dist/src/commands/decode.js:65:41)
    at renderArgs (/usr/lib/node_modules/@usecannon/cli/dist/src/commands/decode.js:45:48)
    at decode (/usr/lib/node_modules/@usecannon/cli/dist/src/commands/decode.js:49:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Command.<anonymous> (/usr/lib/node_modules/@usecannon/cli/dist/src/index.js:335:5)
    ```

**Comment by dbeal-eth** [comment]
the above command is not reproducable everywhere. here is an alterntaive way to reproduce this issue:
1. edit `packages/sample-hardhat-project/contracts/Greeter.sol`. Add method to Greeter:
```
  function takeBytes32(bytes32 arg) public {
    console.log("Got arg");
  }
```
2. Open a terminal in `packages/sample-hardhat-project` and run `npx hardhat cannon:build`
[...]

PR #239, opened 2023-06-14
Handle byte32 parsing issues

aminlatifi:
Ref #233


packages/cli/src/commands/decode.ts
```diff
@@ -75,13 +75,23 @@ function _getAbis(deployData: DeploymentInfo) {
...
+    case type.type === 'address':
+      return value;
+    case type.type === 'bytes32':
+      try {
+        return ethers.utils.parseBytes32String(value as string);
+      } catch (err) {
+        if (process.env.TRACE === 'true') {
+          console.error(err);
+        }
+        return '<ignored>';
```
dbeal-eth, line 90: cuold we instead return the raw, undecoded hex bytes?

dbeal-eth (approved):
could you include a scerenshot of the working new code?

Issue #192, opened 2023-05-09
CLI interact tool should be able to accept "false" for bool inputs


noahlitvin:
It appears to see "false" as a falsey value and doesn't do anything, rather than "false" as a parameter to use

noahlitvin:
This may be resolved already. Please confirm

dbeal-eth:
reproduction example:
1. edit `packages/sample-hardhat-project/contracts/Greeter.sol`. Add method to Greeter:
```
  function takeBoolean(bool arg) public {
    console.log("Got arg", arg);
  }
```
2. Open a terminal in `packages/sample-hardhat-project` and run `npx hardhat cannon:build`
3. Run the command: `node ../cli/dist/bin/cannon.js greeter:2.4.15`. Push `i` to start up the interactive cli
4. navigate to the `takeBoolean` command we just created inside of `Greeter` and type `false` for the arg argument. notice how the behavior is different when passing true vs. false for the command. When passing `false`, the interact CLI exits the input immediately and does not allow for running of the transaction. This should be fixed (ie, you should only exit the function args input if control-c is pushed, in general)

aminlatifi:
Thanks @dbeal-eth for the instructin.

PR #241, opened 2023-06-14
Support falsy values in method arguments

aminlatifi:
Ref #192


dbeal-eth (approved):
good job adding other ways to specify true or false

noahlitvin (approved):

Issue #243, opened 2023-06-16
Customise gas for contract deployment


rolandsaven:
Hey there,

I've been playing around with cannon and trying to build a deployment process, but I'm having trouble customizing the gas values for contract deployment. I'm getting the following error while trying to deploy to Polygon Mainnet. (Deployment to Mumbai worked just fine.)

```json
{
 "code": -32000, 
 "message": "transaction underpriced"
}
```

Regarding reproduction, I tried deploying Synthetix's Oracle manager contract set, but it didn't work. 
https://github.com/Synthetixio/synthetix-v3/blob/main/protocol/oracle-manager/cannonfile.toml 

So if I try to build these contracts on Polygon Mainnet, it fails (Mumbai works).

I've been looking through Cannon's code and see that it's possible to define overrides. 
https://github.com/usecannon/cannon/blob/821b1abddd18232c5c4a84461823a53efb77d296/packages/builder/src/steps/contract.ts#L39-L44

I imagined the following would work, but it does not recognise the values. Am I doing something wrong?

```toml
[contract.InitialProxy]
artifact = "contracts/Proxy.sol:Proxy"
args = ["<%= contracts.CoreModule.address %>", "<%= settings.owner %>"]
salt = "<%= settings.salt %>"
abiOf = ["CoreModule"]
create2 = true
overrides.gasPrice = "150500000000"
overrides.priorityGasPrice = "40000000000"
```

noahlitvin:
Thanks for raising this issue! I think it probably makes sense to make these parameters on the build command rather than part of the cannonfile spec. Thoughts @dbeal-eth @mjlescano @aminlatifi?

dbeal-eth:
Agreed, gas price is a runtime issue. However, gas limit would make more sense on the command itself.

noahlitvin:
Yea makes sense. @aminlatifi can you take this one? (Adding a `--gas-limit` param to the `build` command.)

aminlatifi:
Yep

aminlatifi:
As I see, transaction options (including gas limit, gas price, etc.) aren't supported in the build process. Only the `invoke` action has such a trait which can be inspired from.

dbeal-eth:
> Yea makes sense. @aminlatifi can you take this one? (Adding a `--gas-limit` param to the `build` command.)

should be `--gas-price` not `--gas-limit`, and possibly as a separate ticket, add `gasLimmit` as an option for invoke

PR #255, opened 2023-06-20
Gas Price support to build and interact commands

aminlatifi:
Removed gasPrice and priorityGasPrice from Cannon Config section
Added gas price and fee prices (EIP 1559) support to build and interact commands 
Ref #243


dbeal-eth (changes requested):
looking great!

have we verified that the gas price settings are honored in `invoke` steps as well? have we verified that the gas settings are honored in nested runtimes (aka are settings copied after call to `derive`?)

you might want to check in on my changes here for adding support for "signals" https://github.com/usecannon/cannon/pull/264/files#diff-f3cbb80e150d9b892e9d17220ea87f6e58aee3b540de6b2cff5f4b5749f7939dR71 which may be relevant to this SIP solution.

aminlatifi:
> looking great!
> 
> have we verified that the gas price settings are honored in `invoke` steps as well? have we verified that the gas settings are honored in nested runtimes (aka are settings copied after call to `derive`?)
> 
> you might want to check in on my changes here for adding support for "signals" https://github.com/usecannon/cannon/pull/264/files#diff-f3cbb80e150d9b892e9d17220ea87f6e58aee3b540de6b2cff5f4b5749f7939dR71 which may be relevant to this SIP solution.

@dbeal-eth thanks for the thorough review. I tested whether the gasPrice and others is passed to under layers by logging variables manually at the transaction-creating spots. 

On the `derive` function you are right, there was an issue. I pushed this commit to solve e9fde55d2e8f87cb5dc08dd02bfea87e1caaf431

dbeal-eth (approved):

Issue #20, opened 2022-07-17
Estimate gas for publishing packages


noahlitvin:
Replace "(This will cost a small amount of gas.)" with an actual gas estimate in `publish.ts`.

noahlitvin:
Review the guide for how to publish: https://usecannon.com/get-started

Use the testnet registry (set the registry RPC to goerli and the registry address with `cannon setup`)

mohammadranjbarz:
I couldn't find **This will cost a small amount of gas** text as hard coded message in the repository,  and the only place we get estimated gas price seems is below snippet, do you mean `await contract.estimateGas` return the the text that you said and should I parse the output of that function?

https://github.com/usecannon/cannon/blob/821b1abddd18232c5c4a84461823a53efb77d296/packages/cli/src/interact.ts#L310-L322

noahlitvin:
Ah yea, the codebase has shifted. Instead, let's clean up the CLI output for the publish command in general.

* The overrides parameter includes `maxFeePerGas`, `maxPriorityFeePerGas`, `gasLimit`. (See where this is being called in `cli/src/index.ts`) Let's have the console output start with a output that displays these settings (whether the default or the override) and tells the users which params to use to alter them.
* All error/warning messages should have proper punctuation/capitalization.
* First, the command attempts to the publish the IPFS package to the publishIpfsUrl. Let's output that that's what it's doing "Uploading the following Cannon package data to <publishIpfsUrl>... " with the name, tag, variant, qm hash for each.
* Then, the command executes a transaction on the Cannon registry to associate the name/version/variant with the qm hash. Instead of just spitting out the JSON string at the bottom, let's make it clearer that that's what's happening. Something like "Publishing name:version:variant to the Cannon registry (0x0000...) using signer 0x0000.... on mainnet."

noahlitvin:
Appreciate PR #259. Let's still show the gas estimate.

mohammadranjbarz:
> Appreciate PR #259. Let's still show the gas estimate.

@noahlitvin Can you please give me more detail, in which section do you mean I should log the gas estimate?

FuzzB0t:
Closed with #310

PR #259, opened 2023-06-22
Improve logs on publish

mohammadranjbarz:
related #20

### DIFF
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -60,7 +60,6 @@ export async function publish({
         tags,
         'Variant',
         `${chainId!}-${preset!}`,
-        'QM Hash'
       )
     );
     console.log();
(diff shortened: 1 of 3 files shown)


mohammadranjbarz:
@noahlitvin  It's a little different with what you said in 
https://github.com/usecannon/cannon/issues/20#issuecomment-1596171529

because You said about logging some parameters that I didn't know hot can I have them before publishing package, anyway I think better way is to just create my PR and you leave your feedback here
Sorry I know it's a tiny PR and you might can implement it faster than explaining it

packages/cli/src/commands/publish.ts
```diff
@@ -42,18 +43,27 @@ export async function publish({
...
     console.log(blueBright('publishing remote ipfs package', packageRef));
+    console.log(
+      blueBright(
+        'Uploading the following Cannon package data to',
+        cliSettings.publishIpfsUrl,
+        'Tags',
+        tags,
+        'Variant',
+        `${chainId!}-${preset!}`,
+        'QM Hash'
```
dbeal-eth, line 63: no qm hash is included after

dbeal-eth (approved):
thanks for this

packages/cli/src/commands/publish.ts
mohammadranjbarz: Thanks, Done.

PR #261, opened 2023-06-22
Fix ipfs publish file

mjlescano:
Before, when running the `publish` using an `@ipfs:Qm...` reference, it would only try to publish the given file on the registry, but with this fix it first makes sure to download the tree of artifacts and upload it to the configured `publishIpfsUrl`:
Screenshot 2023-06-22 at 18 04 18


packages/cli/src/index.ts
```diff
@@ -333,7 +333,7 @@ program
 
     await publish({
       packageRef,
-      signer: p.signers[0],
+      signer: p.signers[0] || p.provider,
```
mjlescano, line 336: This is necessary for the case that the user does not configure a signer, so, it will still be able to run Read operations

packages/builder/src/package.ts
```diff
@@ -112,6 +112,48 @@ export async function copyPackage({ packageRef, tags, variant, fromStorage, toSt
...
+  const deployData = await readStorage.readDeploy(packageRef, preset, chainId);
+
+  if (!deployData) {
+    throw new Error('ipfs could not find deployment artifact. please double check your settings, and rebuild your package.');
+  }
+
+  const metaUrl = await readStorage.registry.getMetaUrl(packageRef, `${chainId}-${preset}`);
+
+  const copyIpfs = async (deployInfo: DeploymentInfo) => {
+    const newMiscUrl = await readStorage.putBlob(await readStorage.readBlob(deployInfo.miscUrl));
```
aminlatifi, line 140:
```suggestion
    const newMiscUrl = await publishStorage.putBlob(await readStorage.readBlob(deployInfo.miscUrl));
```
@mjlescano I am not well-versed on this, am I right?

packages/builder/src/package.ts
```diff
@@ -112,6 +112,48 @@ export async function copyPackage({ packageRef, tags, variant, fromStorage, toSt
...
+
+    let newMetaUrl = metaUrl;
```
aminlatifi, line 142: Is this initialization required?

packages/builder/src/package.ts
aminlatifi: Besides, shouldn't we check the upload was successful?

packages/builder/src/package.ts
```diff
@@ -112,6 +112,48 @@ export async function copyPackage({ packageRef, tags, variant, fromStorage, toSt
...
+
+    if (metaUrl) {
+      newMetaUrl = await publishStorage.putBlob(await readStorage.readBlob(metaUrl));
+      if (!newMetaUrl) throw new Error('error while writing new misc blob');
+    }
```
aminlatifi, line 147: Should we save the new MetaUrl somewhere in the copy deployInfo, to upload again?

aminlatifi (changes requested):

mjlescano:
This is not going to be merged, as the `cannon publish` command using an ipfs file funcionality is going to be moved to a new `extract-link` command.

Issue #262, opened 2023-06-22
Parsing / string interpolation in defaultValues for settings


noahlitvin:
Can we have it parse `defaultValue` ?

e.g. a user wants:
```toml
[setting.perpsMaxSnxEth]
defaultValue = "<%= MaxUint256 %>"
```

PR #263, opened 2023-06-25
Settings support in defaultValue

aminlatifi:
load values from settings support in loading default values
Ref #262

### DIFF
--- a/packages/builder/src/definition.ts
+++ b/packages/builder/src/definition.ts
@@ -161,9 +161,11 @@ export class ChainDefinition {
    * @returns definition of settings
    */
   getSettings(ctx: PreChainBuilderContext) {
-    return _.mapValues(this.raw.setting, (s) => {
-      const newSetting = _.clone(s);
-      newSetting.defaultValue = _.template(s.defaultValue)(ctx);
+    const loadedSettings: Record<string, any> = {};
+    return _.mapValues(this.raw.setting, (sValue, sKey) => {
+      const newSetting = _.clone(sValue);
+      newSetting.defaultValue = _.template(sValue.defaultValue)({ ...ctx, settings: loadedSettings });
+      loadedSettings[sKey] = newSetting.defaultValue;
       return newSetting;
     });
   }
--- a/packages/builder/src/definition.ts
+++ b/packages/builder/src/definition.ts
@@ -162,9 +162,10 @@ export class ChainDefinition {
    */
   getSettings(ctx: PreChainBuilderContext) {
     const loadedSettings: Record<string, any> = {};
+    const _ctx = { ...ctx, ...ethers.utils, ...ethers.constants, settings: loadedSettings };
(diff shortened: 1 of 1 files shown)


packages/builder/src/definition.ts
```diff
@@ -161,9 +161,11 @@ export class ChainDefinition {
    * @returns definition of settings
    */
   getSettings(ctx: PreChainBuilderContext) {
-    return _.mapValues(this.raw.setting, (s) => {
-      const newSetting = _.clone(s);
-      newSetting.defaultValue = _.template(s.defaultValue)(ctx);
+    const loadedSettings: Record<string, any> = {};
+    return _.mapValues(this.raw.setting, (sValue, sKey) => {
+      const newSetting = _.clone(sValue);
+      newSetting.defaultValue = _.template(sValue.defaultValue)({ ...ctx, settings: loadedSettings });
```
dbeal-eth, line 167: hey these changes look pretty close! however, this :are a couple of additional complixities that need to be considered before we merge this:
* if you look at `getConfig` (the place where the majority of interpolation handling happens in regular cases), we actually import additional types into the ctx. For example, `ethers.utils.*` is included, as well as `ethers.constants` which includes `MaxUint256` as was given in the issue report.
* obviously when setting up settigns for loading you cant depend on the current setting or any settings that depend on this setting. What restrictions are there and how is it loaded? Unfortunately its not clear here exactly how that works. I would be fine with settings not being able to import other settings as a solution.

ideally, these additional utilities should also be included in any place where we do interpolation, not just in config parsing.

dbeal-eth (changes requested):
i think this is definitely a case where tests need to be added to verify this functionality before this gets merged. as well as some manual testing. see comments! :+1:

packages/builder/src/definition.ts
aminlatifi: @dbeal-eth Good point, I added ethers.utils and ethers.constants support. Regarding supporting in other places there is only one place we load the default values in the code, the same spot I have modified.

However, you have pointed out a more fundamental issue. If I understood well, the order of properties saved in the js doc is not the same as the cannofile, so `defaultValue` of a later variable will be tried to solve first. Did I get it correctly?

dbeal-eth: well, like I said, its "complicated".

the order is *probably* the same, but you have to keep in mind that the JSON document encodes `setting` as a dictionary. This means that when you iterate over all the items in the dictionary, they while there is a good chance the order will be preserved, tis not guarenteed.

many projects ignore this, but I get kind of nervous relying on things that are not required by spec. So thats part of I why I am thinking we should just not support loading any "loadedSettings" or anything like that. AKA, settings cant load from other settings.

aminlatifi: I see, thanks @dbeal-eth

dbeal-eth (approved):

PR #264, opened 2023-06-25
regression and log fixes

dbeal-eth:
new features (and a couple old) had some regressions.

also tighten up logging in some places, minor fixes like that

in particular:
* previously, if executing a build on a gitops repo or anything that uses
nested provisioning. the cancelling of the build will fail because the `cancelled`
boolean only gets applied to the top level runtime, but it needs to be shared between all runtimes

### DIFF
--- a/packages/cli/src/commands/run.ts
+++ b/packages/cli/src/commands/run.ts
@@ -122,7 +122,6 @@ export async function run(packages: PackageSpecification[], options: RunOptions)
       }
 
       buildOutputs.push({ pkg, outputs });
-      provider.artifacts = _.merge(outputs, provider.artifacts);
     }
 
     console.log(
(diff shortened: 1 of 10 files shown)


packages/cli/src/commands/run.ts
```diff
@@ -121,6 +122,7 @@ export async function run(packages: PackageSpecification[], options: RunOptions)
       }
 
       buildOutputs.push({ pkg, outputs });
+      provider.artifacts = _.merge(outputs, provider.artifacts);
```
aminlatifi, line 125: @dbeal-eth I am not pretty sure it won't cause any problem. While till now all scenarios I have run have included a single package, in case we have multiple packages it may cause an issue!
Besides, I have solved the issue here
https://github.com/usecannon/cannon/blob/20327907955ff3ed0fbaaac5f83c86da84dab3bc/packages/cli/src/interact.ts#L87
This was developed before but has been merged after you made this PR.

aminlatifi (changes requested):

mjlescano (approved):

packages/cli/src/commands/run.ts
dbeal-eth: ok as long as you confirm it has been fixed elsewhere I will go head and remove this line.

aminlatifi (approved):

PR #265, opened 2023-06-26
Write tests for cli package

mohammadranjbarz:
related to #260


packages/cli/src/helpers.test.ts
mjlescano, line 1: IMHO, there's no need to test that all the values on a static file are correct, I would just test something like the main 2 or 3 chains (this would check that the file is correctly written and connected to the `getChainId`, `getChainDataFromId` and `getChainName` are working properly), and also test the failing cases.

This is because the tests would just get in the way when we want to update the chains.

dbeal-eth (approved):
great start on the tests! probably couldn't have done them better myself

packages/cli/src/helpers.test.ts
mohammadranjbarz: @mjlescano  I reduced lots of chains and just kept few important chains

mohammadranjbarz:
@dbeal-eth  @mjlescano 
So I merge this PR for other tests ( especially integration tests if we can do it) we can create other PRs later

PR #275, opened 2023-07-02
WIP write tests for alter command

mohammadranjbarz:


packages/cli/src/commands/alter.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    expect(mockedFallBackRegistry.publish as jest.Mock<any, any>).toHaveBeenCalledWith(
+      [packageName],
+      `${chainId}-${preset}`,
+      newUrl,
+      metaUrl
+    );
+  });
+
+  // TODO
+  // it('should perform alteration for mark-complete', async () => {});
```
aminlatifi: @dbeal-eth it looks like the package status must be modified, what do you think?

packages/cli/src/commands/alter.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    const targets = ['provision.dummyStep'];
+
+    // Call the 'alter' function with the necessary arguments
+    await alter(packageName, chainId, preset, testPkgData.meta, command, targets, runtimeOverrides);
+
+    expect(CannonStorage.prototype.readDeploy as jest.Mock<any, any>).toHaveBeenCalledWith(packageName, preset, chainId);
+    expect(CannonStorage.prototype.putDeploy as jest.Mock<any, any>).toHaveBeenCalledWith(testPkgData);
+
+    // TODO: I am not sure the package status must be changed to another value
+    // expect(testPkgData.status).toEqual('incomplete');
```
aminlatifi, line 179: The same

packages/cli/src/commands/alter.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    jest.spyOn(ChainDefinition.prototype, 'getState').mockResolvedValue(hash);
+
+    // Call the 'alter' function with the necessary arguments
+    await alter(packageName, chainId, preset, testPkgData.meta, command, targets, runtimeOverrides);
+
+    expect(CannonStorage.prototype.readDeploy as jest.Mock<any, any>).toHaveBeenCalledWith(packageName, preset, chainId);
+    expect(CannonStorage.prototype.putDeploy as jest.Mock<any, any>).toHaveBeenCalledWith(testPkgData);
+
+    // TODO: I am not sure the package status must be changed to another value
+    // expect(testPkgData.status).toEqual('complete');
```
aminlatifi, line 157: @dbeal-eth it looks like it would be better to modify package.stauts value, WDYT?

packages/cli/src/commands/alter.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    const targets = ['provision.dummyStep'];
+
+    // Call the 'alter' function with the necessary arguments
+    await alter(packageName, chainId, preset, testPkgData.meta, command, targets, runtimeOverrides);
+
+    expect(CannonStorage.prototype.readDeploy as jest.Mock<any, any>).toHaveBeenCalledWith(packageName, preset, chainId);
+    expect(CannonStorage.prototype.putDeploy as jest.Mock<any, any>).toHaveBeenCalledWith(testPkgData);
+
+    // TODO: I am not sure the package status must be changed to another value
+    // expect(testPkgData.status).toEqual('incomplete');
```
aminlatifi, line 179: The same

packages/cli/src/commands/alter.test.ts
```diff
@@ -0,0 +1,225 @@
...
+      newUrl,
+      metaUrl
+    );
+  });
+
+  it('should perform alteration for mark-complete', async () => {
+    const command = 'mark-complete';
+    const targets = ['provision.dummyStep'];
+    const hash = '0xmark-complete-fffffffffffffffffffffffffffffffffffffffffffffffff';
+    jest.spyOn(ChainDefinition.prototype, 'getState').mockResolvedValue(hash);
```
aminlatifi, line 148: I mocked it so generally, should I go deeper and play with actions?

packages/cli/package.json
```diff
@@ -31,7 +31,6 @@
     "ajv": "^8.12.0",
     "axios": "^1.3.3",
     "jest": "^29.5.0",
-    "sinon": "^15.2.0",
```
mjlescano, line 34: ❤️

packages/cli/src/commands/alter.test.ts
dbeal-eth: I don't think it should be altered at all because this value reflects the state of whether the package is complete or not. If after marking complete, the package has all of its steps completes, the user can verify this but running the Canon build command again and it will recognize that there is no steps to run and mark the package is complete

dbeal-eth (approved):
Looking pretty good and I think the alter command is a good place to start with. There's probably a few edge cases there we should write some tests for. I can help figure out what those are if necessary. Overall looking really good.

Issue #268, opened 2023-06-29
Throw error on factory or extra where event is not found


dbeal-eth:
Currently, if a invoke step specifies a extra or factory property which is expecting a particular event , if there is no expected event , Canon will not throw an error or even a warning to indicate to the user that no event was found. This is a common source of errors so I think it would be good to at least default to throwing an error if a matching event is not found and also offer potentially a flag on the invoke step which allows for such errors to be suppressed if you are interacting with a contract which may not emit an event

FuzzB0t:
@dbeal-eth I added warnings for the contract not containing any events and the cannonfile not expecting any events. Also the invoke step now throws a full on error if the cannonfile specifies an event but the invoked function does not emit that event.

The warnings can be suppressed with a new flag I added, but the error cannot be suppressed. Let me know if you have any change requests or feedback!

PR #276, opened 2023-07-02
Invoke event error log

FuzzB0t:


dbeal-eth (changes requested):
It's not necessary to admit a warning when events are admitted, but no event is expected. That is very normal and should not be anything that we need to admit a warning for.

FuzzB0t:
Reference Issue #268

FuzzB0t:
Issue #268

Before when running the build cmd for a cannonfile that contained an invoke step which expected an event from an invoked function, cannon would execute the step without any warning:

image

The above consumer cannonfile expects an event, meanwhile the contract function thats being invoked is not emitting an event, After rebuilding the contract and running build on the consumer cannonfile the output would look like this:

image

FuzzB0t:
After making changes to the invoke step, the build cmd will now throw an error when an event is expected but not emitted. You can choose to ignore these errors by adding the 'ignoreErrors' flag under an invoke step:
image


So by running the build cmd on the context above we now get:

image

FuzzB0t:
Issue #268 

For reference, here's the same test run on the hardhat sample project to test out the changes on the hardhat-plugin:

image

image

FuzzB0t:
Added an example of multiple event emissions in the Greeter contract under the sample foundry contract:

Multiple events get emitted from one TX and the cannonfile uses an event from a previous invoke step to set attributes for a subsequent invoke step:

> Greeter Contracts

image

> Cannonfile.consumer.toml

image

> Build command output log

image

dbeal-eth (approved):
Awesome

FuzzB0t:
Added a few more examples to simply show the multiple events / multiple contract deployments referencing within a cannonfile:

image

image

Also updated the technical docs with info about the extra option in the cannonfile and how to reference events and contracts in subsequent invoke steps.

dbeal-eth (approved):

Issue #281, opened 2023-07-04
Unable to find out/Contract.sol/Contract.json 


avniculae:
Cannon versions > 2.4.14 are unable to find artifact files when running `cannon build cannonfiles/file.toml`. The command is run in the same folder where `out/` and `src/` are; however, note that the cannon file is not in the same directory, but rather in the `cannonfiles` folder.

dbeal-eth:
@avniculae hello! thanks for this report! could you please output the result of `forge config --json` from your project? in particular we are interested in the `out` config option

dbeal-eth:
also you mentioned that your cannonfile was in a sub-directory? does that mean the command you are executing to build your project is in the form `cannon build cannon/package.toml`?

PR #283, opened 2023-07-05
Subdirectory build fail

FuzzB0t:
When using a subdirectory to store your cannonfiles, if the build command was run from the project root directory then cannon would try to fetch artifacts using the subdirectory as the root-directory.

image

image

By using the process.cwd() [Node function](https://nodejs.org/api/process.html#processcwd) we can set the project directory to be relative to where the process is run. 

Here's a successful run after making the change and rebuilding:

image


FuzzB0t:
ref #281

PR #284, opened 2023-07-06
Feat/cli tests inspect command

mohammadranjbarz:


packages/cli/src/commands/inspect.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    expect(mockedFallBackRegistry.getMetaUrl).toHaveBeenCalledWith(`${packageName}`, `${chainId}-${preset}`);
+  });
+  test('should call inspect with json flag ', async () => {
+    // Call the 'inspect' function with the necessary arguments
+    const result = await inspect(packageName, chainId, preset, true, '');
+
+    expect(result).toEqual(testPkgData);
+    expect(mockedFallBackRegistry.getUrl).toHaveBeenCalledWith(`${packageName}`, `${chainId}-${preset}`);
+    expect(mockedFallBackRegistry.getMetaUrl).not.toHaveBeenCalled();
+  });
```
mjlescano, line 86: And, it would be nice to check that the output is being written on `process.stdout`.
I didn't test it, but maybe you can do it listening to the data event like so:
```javascript
testApp.stdout.on('data', data => {
  // assertions here
})
```

dbeal-eth (approved):
Good start

packages/cli/src/commands/inspect.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    // Set up test data and variables
+    const writeDeployments = 'contracts';
+
+    // Call the 'inspect' function with the necessary arguments
+    const result = await inspect(packageName, chainId, preset, false, writeDeployments);
+
+    expect(result).toEqual(testPkgData);
+    expect(mockedFallBackRegistry.getUrl).toHaveBeenCalledWith(`${packageName}`, `${chainId}-${preset}`);
+    expect(mockedFallBackRegistry.getMetaUrl).toHaveBeenCalledWith(`${packageName}`, `${chainId}-${preset}`);
+  });
```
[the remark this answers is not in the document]
mohammadranjbarz: Thanks @mjlescano I added something to test it

packages/cli/src/commands/inspect.test.ts
mohammadranjbarz: @mjlescano Done.

PR #285, opened 2023-07-06
Added cli publish test

aminlatifi:
Ref #260


packages/cli/src/commands/publish.test.ts
```diff
@@ -0,0 +1,232 @@
...
+        recursive: true,
+        overrides: {},
+      });
+
+      expect(copyPackage as jest.Mock).toHaveBeenCalledTimes(2);
+      expect((copyPackage as jest.Mock).mock.calls[0][0].packageRef).toEqual(packageName);
+      expect((copyPackage as jest.Mock).mock.calls[0][0].variant).toEqual(`${chainId}-${preset}`);
+      expect((copyPackage as jest.Mock).mock.calls[1][0].packageRef).toEqual(packageName);
+      expect((copyPackage as jest.Mock).mock.calls[1][0].variant).toEqual(`${otherChainId}-${preset}`);
+    });
```
aminlatifi, line 230: It would be better to remove/skip these tests because they are written based on the current implementation of methods rather than the expected behaviour

packages/cli/src/publish.test.ts
```diff
@@ -0,0 +1,243 @@
...
+      const testPkgDataFilePath = path.join(cliSettings.cannonDirectory, 'tags', deployDataLocalFileName);
+      switch (_path) {
+        case testPkgDataFilePath:
+          return Promise.resolve(Buffer.from(testPkgDataIpfsUrl));
+        case testPkgDataFilePath + '.meta':
+          return Promise.resolve(Buffer.from(testPkgMetaIpfsUrl));
+      }
+      return Promise.resolve(Buffer.from(''));
+    });
+    jest.spyOn(IPFSLoader.prototype, 'read').mockImplementation((url: string): Promise<any> => {
```
dbeal-eth, line 101: Instead of using mocks you could have probably used the in memory registry and loader

dbeal-eth (changes requested):
Sorry for not saying this earlier, but I think these tests would be best implemented and easier to maintain with the `InMemoryLoader` and `InMemoryRegistry` which is a lot easier to set up and is used for prod code

packages/cli/src/publish.test.ts
aminlatifi: @dbeal-eth I couldn't find the `InMemoryLoader` class anywhere in the code, are you sure of the name?
Also, particularly in this case, we don't have control over the type of loader that is instantiated as I see. The `publish` method instantiates one itself, do you mean I mock the whole OnChainRegistery class?

aminlatifi: As I said we don't have control over the type of loader created, or the registry. So I merge this PR.

PR #286, opened 2023-07-06
Fixed issue with custom build path

aminlatifi:
@FuzzB0t Did you have any intention for changing that line? that change causes an issue in building with custom path like `cannon build ../../packages/sample-foundry-project/cannonfile.toml`


noahlitvin:
Closing in favor of #287

PR #287, opened 2023-07-06
Dynamic build path

FuzzB0t:
Ref #286 

Before when running the build step using a path with cannonfiles in a subdirectory it would fail to find artifacts, on the contrary using a relative path would work. I implemented a fix by simply forcing the user to be in the project root directory when running the command so that it would always resolve the parent directory as the foundry project root (therefore being able to find the artifacts).  However this would break the use of relative paths when specifying a cannonfile like '../../packages/sample-foundry-project/cannonfile.toml', I implemented a fix that recursively checks for a 'foundry.toml' file to determine the foundry project root so getArtifacts always has context of the project root directory regardless of where the command is ran and/or where the cannonfiles are (whether in a subdirectory or not).

Before
image

After
image

Before
image
After
image

The alternative to this is to simply force users to run this command within the project root and change the CLI to display a message to the user to rerun the command within the root.

### DIFF
--- a/packages/cli/src/foundry.ts
+++ b/packages/cli/src/foundry.ts
@@ -35,17 +35,17 @@ export async function getFoundryArtifact(name: string, baseDir = ''): Promise<Co
 
     // append 'foundry.toml' to filepath
     const filePath = path.join(currentPath, markerFile);
-  
+
     // If filepath exists it means we're already at root of the project
     // so just return currentPath
     if (fs.existsSync(filePath)) {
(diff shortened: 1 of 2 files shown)


dbeal-eth (approved):

FuzzB0t:
WIP - Needs tweaking, not sure I feel like this is the best option and still may just opt to add a cli message and just assume users run it from the project root, potentially also add something in the docs to clarify that.

PR #305, opened 2023-07-11
Fix cannon executable

noisekit:
1. Bin file must have executable flag `chmod +x`
2. Bin file must not be generated (as when it is generated it will not have executable flag)


dbeal-eth:
nobody has been able to reproduce the error, so I am going to close this PR. if you find yourself running into the issue again, please let us know.

PR #313, opened 2023-07-13
Feat/tests cli run commnd

mohammadranjbarz:
related to #297

### DIFF
--- a/packages/cli/src/commands/run.ts
+++ b/packages/cli/src/commands/run.ts
@@ -137,7 +137,6 @@ export async function run(packages: PackageSpecification[], options: RunOptions)
   }
 
   if (!signers.length) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/run.ts
```diff
@@ -137,6 +137,7 @@ export async function run(packages: PackageSpecification[], options: RunOptions)
   }
 
   if (!signers.length) {
+    //TODO during writing tests I knew we never reach this point, signers is never empty because of this line signers = [provider.getSigner(addr)];
```
noahlitvin, line 140: @mohammadranjbarz Please remove this line and then we'll merge this as-is. Thanks!

mohammadranjbarz: Done

noahlitvin (approved):

Issue #315, opened 2023-07-16
Cannon hangs checking deployment versions while running offline on v2.4.21


FuzzB0t:
this function hangs forever. Needs to fail gracefully and just carry on

```
async function checkCannonVersion(currentVersion) {
    const latestVersion = await execPromise('npm view @usecannon/cli version');
// ... cut
```
 
then even though all packages are built locally and running things offline before had no problems, this error now pops up
```
> cannon run dex:0.0.0
Starting local node...

settings not configured: please create file ~/.local/share/cannon/settings.json for better performance. See https://usecannon.com/docs for more information.
using default settings (https://ipfs.io, https://ethereum.publicnode.com/)
Error: deployment not found: dex:0.0.0. please make sure it exists for the main preset and network 13370
```

and also this occurs trying to build
`Error: deployment not found: mintable-token:latest. please make sure it exists for preset main and network 13370.`

dbeal-eth:
In general for this issue we want a complete cannon build to be able to occur in the absence of an internet connection as long as required package artifacts had already been downloaded

This may also involve auto-caching builds

FuzzB0t:
closed with #323

PR #323, opened 2023-07-31
Build and Run successfully when not connected to internet

mohammadranjbarz:
related to #315

We had two problems:
1. Checking `cannon` version before executing every command, so I changed it to just check `cannon` version if system is not offline
2. After gettingUrl in fallback registry we were calling `onChainRegistry.getUrl` and it throws when we are offline, so I changed it to not call it if system is offline ( it was just calling it to show the log, so not calling that in offline more has no side effects)

I tested  `build` and `run` commands and they are working fast and correct like when it's online

For testing this PR 
1. Go to project directory and enter this `npx lerna run build`
2. Go to `packages/sample-foundry-project` directory
3. Make your system offline (Turning off the wifi)
4. Enter `TRACE=true npx ts-node ../cli/bin/cannon.ts build`
5. Enter `TRACE=true npx ts-node ../cli/bin/cannon.ts run greeter-foundry:2.4.21`

Screen Shot 1402-05-09 at 18 44 43

### DIFF
--- a/packages/cli/src/helpers.ts
+++ b/packages/cli/src/helpers.ts
@@ -10,6 +10,7 @@ import { CANNON_CHAIN_ID, ChainDefinition, RawChainDefinition, ChainBuilderConte
 import { chains } from './chains';
 import { IChainData } from './types';
 import { resolveCliSettings } from './settings';
+import { isConnectedToInternet } from './util/is-connected-to-internet';
 
 export async function setupAnvil(): Promise<void> {
   // TODO Setup anvil using https://github.com/foundry-rs/hardhat/tree/develop/packages/easy-foundryup
@@ -76,6 +77,10 @@ export function execPromise(command: string): Promise<string> {
 }
 
 export async function checkCannonVersion(currentVersion: string): Promise<void> {
+  if (!(await isConnectedToInternet())) {
+    console.log('You are offline so we dont check the latest version of cannon');
+    return;
+  }
   const latestVersion = await execPromise('npm view @usecannon/cli version');
 
   if (currentVersion !== latestVersion) {
--- a/packages/cli/src/registry.ts
+++ b/packages/cli/src/registry.ts
@@ -6,6 +6,7 @@ import { yellowBright } from 'chalk';
 
 import { CliSettings } from './settings';
 import { resolveRegistryProvider } from './util/provider';
+import { isConnectedToInternet } from './util/is-connected-to-internet';
 
 const debug = Debug('cannon:cli:registry');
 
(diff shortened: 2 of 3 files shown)


FuzzB0t (approved):
PR looks good to me.

packages/cli/src/helpers.ts
```diff
@@ -76,6 +77,10 @@ export function execPromise(command: string): Promise<string> {
 }
 
 export async function checkCannonVersion(currentVersion: string): Promise<void> {
+  if (!(await isConnectedToInternet())) {
+    console.log('You are offline so we dont check the latest version of cannon');
```
dbeal-eth, line 81: this can just be a debug message rather than a full on message to the user

packages/cli/src/registry.ts
```diff
@@ -106,8 +107,11 @@ export async function createDefaultReadRegistry(settings: CliSettings): Promise<
         result: string;
         registry: LocalRegistry;
       }) => {
+        if (!(await isConnectedToInternet())) {
+          // When not connected to the internet, we don't want to check the on-chain registry version to not throw an error
+          return;
```
dbeal-eth, line 112: :+1:

packages/cli/src/util/is-connected-to-internet.ts
```diff
@@ -0,0 +1,13 @@
+import { resolve as dnsResolve } from 'dns';
+
+export async function isConnectedToInternet() {
```
dbeal-eth, line 3: can we memoize this result so that it does not try to repeatedly resolve www.google.com after the initial call?

dbeal-eth (changes requested):

packages/cli/src/helpers.ts
mohammadranjbarz: Yeah you are right, done.

packages/cli/src/util/is-connected-to-internet.ts
mohammadranjbarz: Yeah I can do it, but it just takes a few milliseconds, I think it might be good to request when we need it ( I mean maybe we sometimes cache the connected status but after a few seconds later user gets offline, ...)
WDYT?

dbeal-eth (approved):

packages/cli/src/util/is-connected-to-internet.ts
dbeal-eth: part of the problem with ahving it sometimes return true or false (ltes say if the connection is flakey) is that these internal checks that are happening if internet is available, but maybe they intermittently dont happen because the internet is down. so like the warning that you are using a local package might not appear, and this would be bad.

i do realize its probably not even a few milliseconds. Just thinking that in general we check to see if internet is available at the beginning--if it isnt then its like we are in a offline mode

mohammadranjbarz: @dbeal-eth  Yeah you are right, makes sense. I changed it and now it caches the connected to internet status

PR #328, opened 2023-08-03
More build output

noahlitvin:


FuzzB0t (approved):

packages/cli/src/commands/build.ts
```diff
@@ -261,9 +261,15 @@ export async function build({
         )
       );
 
+      console.log(
+        yellow(`Rerunning the same build command will attempt to execute skipped steps. It will not re-run executed steps.`)
```
dbeal-eth, line 265: put in a copy-pastable command to rebuild because it actually isn't the same build command

dbeal-eth (approved):

packages/cli/src/commands/build.ts
FuzzB0t: Just added this to this PR. Should be good to go.

FuzzB0t (approved):
PR looks good to go

PR #333, opened 2023-08-07
read from onchain registry by default

dbeal-eth:
due to the frequent confusion when it comes to regisrty usage + cannon, and cause of bugs when doing real releases  with people having local registry code, default from reading all packages from the remote (aka onchain) registry first.

this is accomplished with a few changes:
* alter the order of the registries in the `createDefaultReadRegistry` function to prioritise onchain registry as long as there is an internet connection and the user hasn't specified otherwise (see below)
* user can override the choice of the default registry by using a setting `registry-priority local`.
* Warn the user with instructions on how to override if their local registry differs from the remote. Instructions are much clearer now that its just a command change that can be used.


FuzzB0t (approved):
Everything looks good, tested this locally and it worked.

Only request I have is to add the new CLI option to the website docs.

dbeal-eth:
> Everything looks good, tested this locally and it worked.
> 
> Only request I have is to add the new CLI option to the website docs.

ah yes that is important, but I had auto merge on so it automatically merged already :(

FuzzB0t:
> > Everything looks good, tested this locally and it worked.
> > Only request I have is to add the new CLI option to the website docs.
> 
> ah yes that is important, but I had auto merge on so it automatically merged already :(

Dont worry about it! Thanks for letting me know. Will add this to my notes of doc changes that need to be made :)

PR #340, opened 2023-08-10
Nextjs front req patch

aminlatifi:
Allow builer handleTxError to accepts both provider or rpc URL
Improved builder package export notation


dbeal-eth (changes requested):
what is the specific need for this? wfhy is it not possible to just do the `new ethers.providers.JsonRpcProvider` wrap from outside handleTxnError?

aminlatifi:
> what is the specific need for this? wfhy is it not possible to just do the `new ethers.providers.JsonRpcProvider` wrap from outside handleTxnError?

We haven't included ethers in frontend dependencies, to reduce bundle size. Nevertheless, I see @noahlitvin added in #341. So if we accept that, we don't need this PR anymore.

PR #349, opened 2023-08-14
fix: package version load on build

aminlatifi:
I have identified that this commit is responsible for the problem https://github.com/usecannon/cannon/commit/f91d19193ac37c2dcd3621c6d12dae6cd8260ad9


dbeal-eth (approved):
good once you verify the original feature (building from stdin) still works

PR #355, opened 2023-08-15
feat(cli): `prune` command

dbeal-eth:
Allows for removal of unused/old artifacts from the cannon storage backends, such as the file or ipfs storage backends.

The following logic is used to determine whether to prune (delete) a cannon package:
* the artifact looks like a cannon package (uses new `generator` field to help with this)
* the artifact reported build date older than `keepAge` (uses new `timestamp` field to help with this)
* the artifact is not included on a list of artifacts which are already on the cannon registry (we never want to delete registry artifacts)
   * registry artifacts are supplied by adding an argument to the command: `--filter-package` which defines which registries your storage is supposed to be retainin artifacts for. If this is a local node, then the timestamp is only considered
 * after the prune pass is completed, inner `miscUrl` artifacts cannot be deleted if they are shared and they are automatically removed from the prune list

the idea is that an ipfs provider could just run htis script once a day cron job or whatever to keep unnecesesary artifacts at a minimum.

regular users of cannon (either with local ipfs or file storage) can just run `cannon prune` to quickly clean out any artifacts they probably dont need.

prune command with no debugging and timer (scans over 1300 artifacts on remote server):

image

tail end of the prune command with debug logging enabled on synthetix cluster:
image

pruning my local registry (normally this would just be `cannon prune`):

image


FuzzB0t:
Honestly this feels like it's good to go just from using the command a few times, I would propose some QOL things like maybe allowing users who run `cannon prune` to have an option to view a list of the artifacts being deleted and let them select which ones they want to or dont want to delete.

This is in case that a user has a lot of artifacts and hasn't been able to sift through them on their own, before selecting to run prune, by giving them a visible list of artifacts they may be better able to filter through the unwanted artifacts and ensure they keep the ones they want. 

Not something I think is necessary, just a proposition. Aside from that LGTM.

dbeal-eth:
> 

yep we can add that later. adding such a tool would likely require adding a descriptio nof the artifact and adding a cli to select/non-select things, both of which we don't have right now. and if we think about the >90% use case, people are just looking to clear out some space on their computer, and probably dont care about micromanaging exacty what gets deleted/not deleted.

looks like I got some tests to fix

FuzzB0t:
I think your test will pass if you merge main into your branch

FuzzB0t:
> > 
> 
> yep we can add that later. adding such a tool would likely require adding a descriptio nof the artifact and adding a cli to select/non-select things, both of which we don't have right now. and if we think about the >90% use case, people are just looking to clear out some space on their computer, and probably dont care about micromanaging exacty what gets deleted/not deleted.
> 
> looks like I got some tests to fix

Makes sense, I would say this ticket is ready to go then once the tests are fixed.

PR #364, opened 2023-08-25
fix: remove anvil from inspect command and add support for target string

FuzzB0t:


packages/builder/src/schemas.zod.ts
```diff
@@ -24,6 +24,21 @@ const jsonAbiPathRegex = RegExp(/^(?!.*\.d?$).*\.json?$/, 'i');
...
+// Invoke target string schema
+const targetString = z.string().refine(
+  (val) =>
+    ethers.utils.isAddress(val) ||
+    !!val.match(interpolatedRegex) ||
+    !!val.match(stepRegex) ||
+    !!val.match(artifactNameRegex) ||
+    !!val.match(artifactPathRegex),
+  (val) => ({
+    message: `"${val}" must be a valid ethereum address, previously defined contract step name, contract artifact name or filepath`,
```
dbeal-eth, line 36: `previously defined contract step name` => `existing contract step name`

its not necessary that the contract step was "previous" (although it must be depended upon) (soon we are  getting rid of `depends` so probably just best to keep response simple)

dbeal-eth (changes requested):
PR title does not match contents

PR appears to be making change to allow for target to be an array or non-array, but the title states somethign completely different which isnt included.

FuzzB0t:
> PR title does not match contents
> 
> PR appears to be making change to allow for target to be an array or non-array, but the title states somethign completely different which isnt included.

Yeah this PR contains both changes, although somewhat irrelevant to eachother, I'll change the PR title to match the contents better.

dbeal-eth (approved):

PR #406, opened 2023-09-05
feat(router): Include router action from cannon-plugin-router

mjlescano:
Closes https://github.com/usecannon/cannon/issues/397

Include the plugin logic from https://github.com/Synthetixio/cannon-plugin-router on the core library.

This PR also takes care of removing the plugin if it's installed on the users' machine so it doesn't cause any conflicts.


packages/builder/src/schemas.zod.ts
```diff
@@ -457,3 +457,10 @@ export const chainDefinitionSchema = z
       })
       .deepPartial()
   );
+
+export const routerSchema = z.object({
```
dbeal-eth, line 461: oh wow I just realized we arent just installing the router by default; we are embedding it into the builder? I guess thats fine, it is a super powerful action that probably deserves a standard of its own

dbeal-eth (approved):

PR #409, opened 2023-09-06
feat(website): Add Router Guide

noahlitvin:


packages/website/src/features/Router/RouterPage.tsx
```diff
@@ -0,0 +1,220 @@
...
+          borderBottomColor="gray.800"
+          letterSpacing="0.2px"
+        >
+          Use a Synthetix Router
+        </Heading>
+        <Text mb={4}>
+          There is a limit to the size of smart contracts deployed on EVM
+          blockchains. This can create complications during the development of
+          protocols, where engineers may want an arbitrary amount of code to be
+          executable at a single address.
```
dbeal-eth, line 81: additionally, its difficult to maintain large code files/inheritence trees, which single contract-based development leads to

packages/website/src/features/Router/RouterPage.tsx
```diff
@@ -0,0 +1,220 @@
...
+        </Text>
+        <Text mb={4}>
+          In this guide, we’ll walk through{' '}
+          <Link
+            isExternal
+            href="https://github.com/usecannon/cannon/tree/main/examples/router-architecture"
+          >
+            a simple example
+          </Link>{' '}
+          that uses the router and adds a transparent upgradable proxy.
```
dbeal-eth, line 103: would also be good to link somewhere to ale's original design document on notion somewhere on here, or at least explain more technically how the router actually works. this guide focuses on explaining how to use it in practice obviously.

packages/website/src/features/Router/RouterPage.tsx
```diff
@@ -0,0 +1,220 @@
...
+          Create a Router
+        </Heading>
+        <Text mb={4}>Start by installing/upgrading Cannon:</Text>
+        <Box mb={4}>
+          <CommandPreview command="npm i -g @usecannon/cli" />
+        </Box>
+        <Text mb={4}>
+          Also make sure you have an{' '}
+          <Link isExternal href="https://docs.ipfs.tech/install/ipfs-desktop/">
+            IPFS node
```
dbeal-eth, line 116: technically this is optional now. we have the ability to store objects in local files. it just means that you will be dependent upon public registries for downloading packages, however.

packages/website/src/features/Router/RouterPage.tsx
```diff
@@ -0,0 +1,220 @@
...
+            the proxy contract
+          </Link>
+          :
+        </Text>
+        <Box mb={4}>
+          <CodePreview code={code2} language="toml" />
+        </Box>
+        <Text mb={4}>
+          If you alter one of your contracts, when building, Cannon will
+          automatically detect this, generate a new router, and upgrade the
```
dbeal-eth, line 189: might also want to point out that any untouched contracts on the router will be recycled, saving precious gas. so you dont have to worry about the details of your deployment/what has changed. just focus on writing your changes, running deployments as necessary, and watch out for collisions.

packages/website/src/features/Router/RouterPage.tsx
```diff
@@ -0,0 +1,220 @@
...
+          </Link>
+          , you can use the{' '}
+          <Link as={NextLink} href="/deploy">
+            deployer
+          </Link>{' '}
+          to run upgrades. When your protocol no longer needs to be upgraded, it
+          can be made immutable with a call to <Code>renounceOwnership</Code> on
+          the proxy.
+        </Text>
+      </Box>
```
dbeal-eth, line 217: might want to add a section discussing how to handle storage collisions. Obviously storage collisions are outside the scope of what cannon does for you, but for any upgradable contract architecture it shouldn't hurt to add a quick appendix or somtething with useful guides related to running proxies--including avoiding collisions.

dbeal-eth (changes requested):
in the future could we refer to this guide with something a little more detailed than "guide work"? for example, "add router guide"

PR #415, opened 2023-09-07
fix(builder): Add preset reference in source/package name

FuzzB0t:


packages/builder/src/steps/import.ts
```diff
@@ -33,8 +33,18 @@ export default {
   async getState(runtime: ChainBuilderRuntime, ctx: ChainBuilderContextWithHelpers, config: Config) {
     const cfg = this.configInject(ctx, config);
 
-    const preset = config.preset ?? 'main';
```
FuzzB0t, line 36: This conditional wasnt necessary as this was already being done in the configInject, so long as the cfg variable was used then it would default to 'main' if the defined preset was invalid or nonexistent

packages/builder/src/steps/import.ts
```diff
@@ -62,6 +72,16 @@ export default {
     const importLabel = packageState.currentLabel?.split('.')[1] || '';
     debug('exec', config);
 
+    /*
+     * Source preset reference
+     * Checks for preset reference in source name, anything after the '@' references a preset.
+     * If a preset reference is found in the package name by default it will take preference over a defined preset.
+     */
+    if (config.source.includes('@')) {
```
dbeal-eth, line 80: might make sense to turn the package name into a proper data structure at this point, rather than having this splitting/math happening all over the place

dbeal-eth (changes requested):
minor thoughts. maybe time to refactor package name into a real data structure/class that takes in the string and auto splits/validates it for you?

packages/builder/src/package.ts
```diff
@@ -15,6 +15,70 @@ export type CopyPackageOpts = {
...
+  name: string;
+  version: string;
+  preset: string;
+  formatted(): string;
+}
+
+/**
+ * Used to format any reference to a cannon package and split it into it's core parts
+ */
+export class PackageReference implements CannonPackageReference {
```
FuzzB0t, line 28: @dbeal-eth I added this package reference class and added it to provision and import steps for parsing out name, version and preset from package sources.

packages/builder/src/steps/import.ts
```diff
@@ -47,8 +50,19 @@ export default {
   configInject(ctx: ChainBuilderContextWithHelpers, config: Config) {
     config = _.cloneDeep(config);
 
-    config.source = _.template(config.source)(ctx);
-    config.preset = _.template(config.preset)(ctx) || 'main';
+    
+    const packageRef = new PackageReference(_.template(config.source)(ctx));
+
+    // If both definitions of a preset exist, its a user error.
+    if (config.preset && packageRef.preset) {
```
FuzzB0t, line 57: Need to rewrite this, this check is redundant since packageRef.preset always defaults to main.

dbeal-eth (approved):

PR #424, opened 2023-09-10
fix: misc debug mode fixes

dbeal-eth:
some minor bugs when debugging cannon came up, figured I would just fix these here. fixes are:

* add `registryPriority` option to `hardhat-cannon` cli
* allow for user to press control-c multiple times to force cancel cannon
* remove signal handler for control-c press once build has completed. otherwise cannon will awkwardly block cancelling
* add sanity check to buildLayer function in case a step depends on itself, that wont (step can still depend on itself indirectly but this one case seems to prevent a lot of infinite loops)


FuzzB0t (approved):
LGTM, just needs test fixed and it should be good to go.

Issue #387, opened 2023-08-30
Registry Upgrade


noahlitvin:
* Charge 1 wei (configurable) via msg.value on namespace registration and package publish, revert with `FeeRequired(uint256 amount)`
* Something around renouncing namespace ownership and unpublishing packages 
* Decide on verification being on package in addition to namespace, etc. Configurable 1 wei fee here as well?

PR #425, opened 2023-09-10
feat(fee): Add package publication fee of 1 wei

mjlescano:
https://github.com/usecannon/cannon/issues/387


dbeal-eth (commented):
Lot of other changes on this project besides the 1 wei fee. Also is there a function in the cannon registry for the owner to withdraw the ETH tokens?

Issue #369, opened 2023-08-30
Add ability to pass params through to anvil


noahlitvin:
e.g. `anvil --hardfork paris`

PR #445, opened 2023-09-13
feat(cli): add anvil custom options to run and build commands

mjlescano:
Closes https://github.com/usecannon/cannon/issues/369


dbeal-eth (approved):
Have we been able to test this on synthetics deployments or something like that?

PR #453, opened 2023-09-14
fix(website): fix chains modal overflow

### LINKED ISSUE #426: Network list is weird/too long
There are too many options in the network selection ui from rainbowkit. Can we adjust the UI, at least make it easy to find more common networks?

0xDmitry:
Closes #426

@noahlitvin I've fixed overflow issue, but there are too many options still. Looks like there is no way to add a search or something like that, so the only way is to specify chain list manually (right now we use all supported chains). Do you have some ideas about relevant chain list that we want to use? Or I can reduce it myself and we will add missing ones ad hoc.

Screenshot 2023-09-15 at 16 55 58

### DIFF
--- a/packages/website/src/app/walletProvider.tsx
+++ b/packages/website/src/app/walletProvider.tsx
@@ -44,9 +44,22 @@ const wagmiConfig = createConfig({
 });
 
 function WalletProvider({ children }: { children: ReactNode }) {
+  // NOTE: have to hack the style below becuase otherwise it overflows the page.
+  // hopefully the class name doesnt change from compile to compile lol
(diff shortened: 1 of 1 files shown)


dbeal-eth:
@0xDmitry could oyu share quick screenshots of the resulting work here in the PR in the future when you can? thanks!

dbeal-eth (approved):
lol indeed

0xDmitry:
> @0xDmitry could oyu share quick screenshots of the resulting work here in the PR in the future when you can? thanks!

Sure, np! Will add for this one tomorrow

PR #455, opened 2023-09-14
feat(cli): add `cannon trace` command

dbeal-eth:
implements `cannon trace`, which allows for the generation of a **stack
trace** of *any* transaction on chain, or a specified transaction code
offchain.

the cannon trace command works similar to the `cannon decode` command.
Simply supply a package name and hex string.

can be used to decode any transaction already on chain by submitting the
transaction has as the hex string. The exact parameters under which the
original transaction executed will be accomodated in the simulation to
ensure that the result is exactly the same as its original circumstance
(individual fields, such as `from` or `block`, can be overridden if
desired)

alternatively, if the transaction isnt submitted on-chain or you want to
trace a *read-only* call, supply the exact call data of the transaction
you want to execute. Cannon will auto-detect the contract to send the
call to and submit the transaction automatically. To override the
destination contract (as well as other options such as `value` or
`from`), those values can be supplied as needed.


dbeal-eth:
a good sample transaction is `$CANNON trace --chain-id 10 synthetix-omnibus:3.2.0 0x82f57c8019d760ddca14f1c69bb3fbeb21ba09c7fcd01a13d84a686524ef1e0f`

image

dbeal-eth:
a good example of simulating a transaction which is not on chain (ex. a view function call) is `node packages/cli/dist/bin/cannon.js trace --chain-id 10 synthetix-omnibus 0x2fb8ff2400000000000000000000000000000000000000000000000000000000000000010000000000000000000000008700daec35af8ff88c16bdf0418774cb3d7599b4 -t 0xffffffaEff0B96Ea8e4f94b2253f31abdD875847`

image

dbeal-eth:
@noahlitvin wondering if it would be better to make the transaction output look more colorful?

PR #472, opened 2023-09-21
fix: fix preset being added to deploy local tag filename

FuzzB0t:


packages/builder/src/package.ts
```diff
@@ -137,7 +137,8 @@ export async function copyPackage({ packageRef, tags, variant, fromStorage, toSt
     const preCtx = await createInitialContext(def, deployInfo.meta, 0, deployInfo.options);
 
     return {
-      packagesNames: [def.getVersion(preCtx), ...(context ? context.tags || [] : tags)].map(
+      // TODO (FIX): When using an interpolated <%= package.version %>, def.getVersion doesnt return a value properly.
+      packagesNames: [def.getVersion(preCtx) || version, ...(context ? context.tags || [] : tags)].map(
```
FuzzB0t, line 141: When publishing a package that has the interpolated package value like `<%= package.version %>`, `def.getVersion` fails to parse out the value from it and I'm not sure what's causing it. @mjlescano If you dont mind taking a look at this that would be great.

packages/cli/src/commands/build.ts
```diff
@@ -278,7 +278,7 @@ export async function build({
 
     if (persist) {
       await resolver.publish(
-        [`${name}:latest`, `${name}:${version}@${selectedPreset}`],
```
FuzzB0t, line 281: This line caused a bug where the preset is being added to the filename in the tags directory, causing the scanDeploy / readDeploy functions to be unable to find the package deploy data

PR #479, opened 2023-09-25
feat: Add fetch command and improve publish outputs

FuzzB0t:
Overview of changes in this branch:

- [x] Add Fetch command + tests
- [x] Improve Publish CLI outputs 
- [x] Improve Publish command error handling on estimateGas
- [x] Fix provider resolution when running local node
- [x] add confirmation prompt to publish 
- [x] replace `recursive` option with `--include-provisioned` option to determine whether to publish provisioned packages
 - Might need to add a recursive option where we can do publishMany without including provisioned packages so we can save on tx calls when a user wants to publish a lot of packages at once without including provisioned packages.
- [x] add package selection prompt when a specific version is not provided in the reference

Things that could be improved:
- Automatically trace tx revert tracing.  (for now people can use cannon trace to determine why a publish tx may have been reverted)
- Add better tests for the publish command
- recursively look for and fetch provisioned packages when fetching 

~Note: When testing on actual chains (like goerli) the estimateGas function fails with (execution reverted 0x) and the --include-provisioned option fails on Goerli unless passed a specified gas-limit (What worked for me was gas-limit over 70,000)~

image

~Running trace on this just returns a 0x so Im guessing there's a revert happening when trying to estimateGas for publish. I had ran into an error where the address I used wasnt authorized so I was getting 0x reverts on the actual tx but once I fixed that it went away.~

~this succeeds on the first publish call and fails after the first publish call when connected to a local node.
On Goerli this fails from the start.~

The above issues have been fixed.

Note: It might be worth adding trace to the publish tx to understand when reverts happen and why, like when a Unauthorized() revert happens its hard to tell.


FuzzB0t:
New publish output looks something like this:
image

packages/cli/src/commands/build.ts
```diff
@@ -261,11 +261,15 @@ export async function build({
   // save the state to ipfs
   const miscUrl = await runtime.recordMisc();
 
+  const chainDef = def.toJson();
+
+  chainDef.version = chainDef.version.startsWith('<%=') ? pkgVersion : chainDef.version;
```
FuzzB0t, line 266: I added this line to store chain definition version with interpolated value, before this when trying to publish a package with a interpolated version it had no way of knowing what that packages version is without it being in the chain definition

examples/sample-foundry-project/cannonfile.consumer.toml
```diff
@@ -74,9 +74,7 @@ target = ["greeters.greeter"]
 func = "changeCloneGreeting"
 args = ["<%= extras.NewClonedGreeterEvent_1 %>", "<%= formatBytes32String('Newest Greeting!') %>"]
 
-
-
-[run.empty]
-exec = "./scripts/deploy.js"
-func = "empty"
-modified = ["dummy.js"]
+# [run.empty]
```
dbeal-eth, line 77: probably should jsut remove this entirely

examples/sample-foundry-project/scripts/deploy.js
dbeal-eth, line 1: new empty file?

packages/builder/src/builder.ts
```diff
@@ -31,8 +31,12 @@ export async function createInitialContext(
     chainId,
   };
 
+  debug('define chain builder context');
+
   const settings: ChainBuilderContext['settings'] = {};
 
+  debug('get settings', preCtx);
```
dbeal-eth, line 38: careful is this going to print out anything sensitive?

packages/builder/src/package.ts
```diff
@@ -100,9 +101,52 @@ function _deployImports(deployInfo: DeploymentInfo) {
...
+  const getPackages = async (deployInfo: DeploymentInfo, context: BundledOutput | null) => {
+    debug('create chain definition');
+
+    const def = new ChainDefinition(deployInfo.def);
+
+    debug('create initial ctx with deploy info', deployInfo);
+
+    const preCtx = await createInitialContext(def, deployInfo.meta, deployInfo.chainId!, deployInfo.options);
+
+    debug('created initial ctx with deploy info');
```
dbeal-eth, line 126: debugs here are probably a bit excessive

packages/cli/src/commands/build.ts
```diff
@@ -274,11 +278,14 @@ export async function build({
...
-    const metaUrl = await runtime.putBlob(await readMetadataCache(`${pkgName}:${pkgVersion}`));
+    const metadata = await readMetadataCache(`${pkgName}:${pkgVersion}`);
+
+    const metaUrl = await runtime.putBlob(metadata);
 
+    // locally store cannon packages (version + latest)
     if (persist) {
       await resolver.publish(
-        [`${name}:latest`, `${name}:${version}`],
+        [`${name}:${version}`, `${name}:latest`],
```
dbeal-eth, line 288: ooc what was the reason for switching these here?

packages/cli/src/commands/build.ts
```diff
@@ -314,7 +321,13 @@ export async function build({
 
       console.log(yellow('Run ' + bold(`cannon publish ${deployUrl}`) + ' to pin the partial deployment package on IPFS.'));
     } else {
-      console.log(greenBright(`Successfully built package ${bold(`${name}:${version}@${selectedPreset}`)} (${deployUrl})`));
+      console.log(
+        greenBright(
+          `Successfully built package ${bold(
+            `${name}:${version}@${selectedPreset}`
+          )} \n - Deploy Url: ${deployUrl} \n - Metadata Url: ${metaUrl}`
```
dbeal-eth, line 328: displaying the metadata url here might cause confusion for those who dont understand what it is, having 2 ipfs urls. I think it would be best if we only display the deploy url

packages/cli/src/commands/fetch.test.ts
```diff
@@ -0,0 +1,190 @@
...
+afterEach(() => {
+  mockfs.restore();
+});
+
+jest.mock('../registry');
+jest.mock('../settings');
+jest.mock('../loader');
+jest.mock('../helpers');
+
+describe('fetch', () => {
```
dbeal-eth, line 24: awesome tests :+1:

packages/cli/src/index.ts
```diff
@@ -359,20 +373,47 @@ program
...
+
+    if (!options.privateKey && !process.env.PRIVATE_KEY) {
+      const keyPrompt = await prompts({
+        type: 'text',
+        name: 'value',
+        message: 'Please provide a Private Key',
+        style: 'password',
+      });
+
+      if (!keyPrompt.value) {
```
dbeal-eth, line 404: can we do better verification here? for example, check to make suer the value is 66 characters long and is a hex string starting with 0x, or 64 characters long and a hex string *not* starting with 0x. ethers probably has a function to help with this.

dbeal-eth (changes requested):

packages/builder/src/builder.ts
FuzzB0t: Yeah this debug shouldnt be here actually, was mostly using it to debug an issue.

packages/cli/src/commands/build.ts
FuzzB0t: Sounds good! The reason I displayed it is because I added functionality to the fetch command to include metadata. Not sure if thats the right way to go about it.

packages/cli/src/index.ts
FuzzB0t: Absolutely, good catch. Thanks for bringing this up.

packages/cli/src/util/provider.ts
```diff
@@ -70,27 +70,53 @@ export async function resolveProviderAndSigners({
...
+  const localProviders = {
+    frame: 'http://localhost:1248',
+    direct: 'http://localhost:8545',
+  };
+
+  let ethersProvider;
+  const providerList = [];
+
+  // force provider to use JSON-RPC instead of Web3Provider for local instances
+  if (checkProviders.includes('frame') || checkProviders[0].startsWith('http://')) {
```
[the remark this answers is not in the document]
FuzzB0t: This is actually somewhat wrong still, I dont think we want to setup a JSON-RPC provider for frame EVERYTIME its used, before I was defaulting to that . I think we only want that when connecting to an http endpoint directly so I believe it should just be the this:

```
 // force provider to use JSON-RPC instead of Web3Provider for local instances
  if (checkProviders[0].startsWith('http://')) {
    ethersProvider = new ethers.providers.JsonRpcProvider;
```

PR #480, opened 2023-09-27
fix(cli): foundry doesn't resolve evm version

dbeal-eth:
leads to problems on etherscan verification/reproduction of correct bytecodes


FuzzB0t (approved):

FuzzB0t:
@dbeal-eth is this ready to go? Wondering if I can add this to the upcoming release

PR #489, opened 2023-10-04
chore: Update deps

noisekit:
- [x] fix few missing deps (not comprehensive check with depcheck but just to fix some compilation problems)
- [x] configure typechain to target ethers v5 explicitly
- [x] update deps (except ethers). Some deps moved to use esm, so also skip
- [x] fix for build
- [ ] fix for tests:
  - [x] jest build for esm/ts (problem was with `untildify` switched to esm)

    ```sh
    FAIL  src/helpers.test.ts
    ● Test suite failed to run
       
    Jest encountered an unexpected token
       
    Jest failed to parse a file. This happens e.g. when your code 
    or its dependencies use non-standard JavaScript syntax,
    or when Jest is not configured to support such syntax.
       
    Out of the box Jest supports Babel, which will be used to transform your files
    into valid JS based on your Babel configuration.
    ```

  - [ ] evm incompatibility for router

    ```sh
    FAIL  src/steps/router.test.ts (14.83 s)
    ● steps/router.ts › exec() › generates and deploys Router with a single contract
       
    There was an error when compiling "Router". Invalid EVM version requested.
    ```

  - [x] `cannon-registry:test` (problem was with `untildify` switched to esm)

    ```sh
    Error HH19: Your project is an ESM project (you have "type": "module" set in your package.json) 
    but your Hardhat config file uses the .js extension
    ```

- [x] make sure all deps use `^` versions and not lock version number
- [x] dedupe lock/deps
- [x] remove prettier plugins (unnecessary, slow and also failing) and replace with `prettier --check .`

build should be working well (with secrets)

image


noahlitvin:
Since v2.11 we've made some major changes to the dependency structure. We'll discuss internally better systems for dependency management, but closing this for now. Appreciate the PR!

PR #497, opened 2023-10-11
fix: incorrect handling of providerUrl

dbeal-eth:
for some reason the eth-provider library does not properly handle explicit `--provider-url` when they are specified, at least for localhost forks which are created for a manual test. When this occurs, it becomes impossible to specify a local fork as a target network for a test. this error probably only is relevant when the user is running a foundry instance (and goes away as soon as the foundry instance is shut down).

To solve this problem, bypass eth-provider ocmpletely when an explicit url is provided by the user.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -44,6 +44,162 @@
         "node": ">=0.10.0"
       }
     },
+    "node_modules/@achingbrain/ip-address": {
+      "version": "8.1.0",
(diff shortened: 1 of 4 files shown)


FuzzB0t:
Did something similar to this in my current PR cause I noticed it was breaking. I tested it pretty thoroughly and didnt see any negative side effects.

https://github.com/usecannon/cannon/pull/479/files#diff-f1562640515f07b9bd13ef2532e931c233355aa83a372f0a194f7b20c7c183c8

FuzzB0t (approved):

PR #505, opened 2023-10-14
feat(cli): implement alter `import` command

dbeal-eth:
Allows for a very terraform-like import capability to pull the state of a package item from an on-chain resource.

more info on the terraform import command: https://developer.hashicorp.com/terraform/cli/commands/import

this is achieved by adding a new optional implementation function on the cannon actions, `importExisting`, which allows for specification of a string (such as a transaction hash or contract address, whatever will lead to the ultimate resource that cannon can pull off of). Cannon can then recreate the state of the package from this information.

for now, I am just implementing this feature for the invoke command, since that is the one that most badly needs this capabilty. however, it might be useful to add this to other steps later.

implementation for https://linear.app/usecannon/issue/CAN-73

need to write tests and do some practical testing


FuzzB0t (approved):
Looks good, tested it out locally and didnt run into any issues + don't see any obvious issues at a glance from looking at the code.

Oddly enough the cli CI test is failing but not sure what's causing it. Might need to merge main and fix conflicts but aside from that, lgtm.

PR #508, opened 2023-10-15
fix(cli): much faster startup

dbeal-eth:
often cannon cli would startup slow because the `npm view` command that is called to resolve the latest cannon version has to call the internet to get the latest version. Its not necessary to call out to the internet every time the user uses cannon to check for the latest version, just very periodically.

to solve this problem, the `npm view` command is called only once a week at max. in other cases, the version is loaded from a file stored in their cannon directory.

on my computer this brings the startup time of the cannon process from 1-2 seconds to start every time to near instant.

### DIFF
--- a/packages/cli/bin/cannon.js
+++ b/packages/cli/bin/cannon.js
@@ -3,7 +3,7 @@
 const { red } = require('chalk');
 const cli = require('../dist/src');
 
-cli
+cli.default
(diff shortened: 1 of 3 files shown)


FuzzB0t (approved):

dbeal-eth:
do not merge, needs mkdirp

PR #509, opened 2023-10-16
fix(cli): proper error on providerurl chainid and chainid mismatch

StuartGavidia:
When a user provides a providerurl and chainid, ensure that the chain ids match else display a proper error message

Screenshot 2023-10-24 at 9 27 27 AM

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -11,7 +11,7 @@ import {
   CannonStorage,
 } from '@usecannon/builder';
 
-import { checkCannonVersion, loadCannonfile } from './helpers';
+import { checkCannonVersion, loadCannonfile, getChainIdFromProviderUrl } from './helpers';
 import { parsePackageArguments, parsePackagesArguments, parseSettings } from './util/params';
 
 import pkg from '../package.json';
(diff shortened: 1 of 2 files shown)


packages/cli/src/index.ts
```diff
@@ -106,6 +106,15 @@ function configureRun(program: Command) {
 
       options.port = Number.parseInt(options.port) || 8545;
 
+      if (options.chainId && options.providerUrl) {
+        const providerChainId = await getChainIdFromProviderUrl(options.providerUrl);
+        if (providerChainId != options.chainId) {
+          throw new Error(
+            `Provided providerUrl's blockchain chainId ${providerChainId} does not match with chainId you provided ${options.chainId}`
```
dbeal-eth, line 113: `Supplied providerUrl's blockchain chainId...`

dbeal-eth (approved):
please share a picture of what it looks like when its working (ie when you put in a provider differing from the chain id)

packages/cli/src/index.ts
StuartGavidia: Adding screenshots

PR #513, opened 2023-10-17
fix(website): fix contract nav highlighting

0xDmitry:
Screenshot 2023-10-17 at 21 44 05

### DIFF
--- a/packages/website/src/features/Packages/Interact.tsx
+++ b/packages/website/src/features/Packages/Interact.tsx
@@ -21,15 +21,16 @@ export const Interact: FC<{
   name: string;
   tag: string;
   variant: string;
+  moduleName: string;
+  contractName: string;
   contractAddress: string;
-}> = ({ name, tag, variant, contractAddress }) => {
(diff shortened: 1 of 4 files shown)


packages/website/src/features/Packages/Interact.tsx
```diff
@@ -21,15 +21,16 @@ export const Interact: FC<{
   name: string;
   tag: string;
   variant: string;
+  moduleName: string;
```
dbeal-eth, line 24: Maybe just call this file name?

packages/website/src/features/Packages/Tabs/InteractTab.tsx
```diff
@@ -38,7 +38,19 @@ export const InteractTab: FC<{
   });
 
   const pathName = usePathname();
-  const activeContract = pathName.split('interact/')[1];
+
+  let activeContractOption: Option | undefined;
```
dbeal-eth, line 42: I've worked in other web frameworks in the past where it wasn't necessary to manually split the path information. I'm wondering if that's also the case here?

noahlitvin (approved):

PR #529, opened 2023-10-23
fix(builder): handling of package url modification during copy

dbeal-eth:
when copying from local fs (aka file:// scheme), it is possible that an uploaded package can end up not having its url updated. this leads to builder issues because `file://` scheme artifacts generally only exist on the builder's system, so we have to make sure these get scrubbed out always.

to fix, the `forPackageRecursive` has been updated to automatically update the url stored inside the deployment info object if a url is returned by the copy function. for functions which do not return a url, this change has no effect.

### DIFF
--- a/packages/builder/src/package.ts
+++ b/packages/builder/src/package.ts
@@ -1,4 +1,5 @@
 import Debug from 'debug';
+import _ from 'lodash';
 import { BundledOutput, DeploymentInfo } from './types';
 import { ChainDefinition } from './definition';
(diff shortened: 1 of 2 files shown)


dbeal-eth:
this change was tested by using the `sample-hardhat-project`, building the greeter and the consumer sample packages with no `ipfsUrl` set in the cannon settings (this results in artifacts being stored on the local fs). After that, i call `cannon publish` and verify that the uploaded artifacts do not contain `file://` url

PR #531, opened 2023-10-23
fix(cli): make clean remove all files/folders in tags and metadata_cache

StuartGavidia:
Clearing out all files and folders in tags and metadata_cache. Ensuring their deletion


FuzzB0t (approved):
This looks good just need to fix lint issues and PR commit.

dbeal-eth (approved):
please check CI failures before merging

PR #532, opened 2023-10-23
fix(cli): add inspect enhancements

StuartGavidia:
Inspect output without --sources
Screenshot 2023-10-22 at 10 33 08 PM

Inspect output with --sources
Screenshot 2023-10-22 at 10 33 59 PM


FuzzB0t (approved):
This looks great, overall great work on the tests and adding proper comments to your code. Once you get the CI to pass this should be good to go.

packages/builder/src/types.ts
```diff
@@ -44,6 +44,14 @@ export type ContractData = {
   highlight?: boolean;
 };
 
+export type ArtifactContracts = {
```
dbeal-eth, line 47: something like

`export type ArtifactContracts = { artifacts: Pick<ChainArtifacts, 'contracts'> };`

should work

also, is the `artifacts` wrapper really necessary?

packages/cli/src/commands/inspect.ts
```diff
@@ -91,9 +98,42 @@ export async function inspect(
...
         .join(' ') || '(none)'
     );
+    packageOwner ? console.log('           Owner:', packageOwner) : console.log('          Source:', localSource || '(none)');
     console.log('     Package URL:', deployUrl);
     console.log('        Misc URL:', deployData.miscUrl);
-    console.log('Package Info URL:', metaUrl);
+    console.log('Package Info URL:', metaUrl || '(none)');
+    console.log('Cannon generator:', deployData.generator);
+    console.log('       timestamp:', deployData.timestamp);
+    console.log('IPFS Availability Score(# of nodes): ', ipfsAvailabilityScore || 'Run IPFS Locally to get this score');
```
dbeal-eth, line 107: since this new field is longer than all the other ones... im torn on if we should align all of them so that it doesn't look lik ethe availability store is awkwardly sticking out?

maybe add a line in between, like:

```
console.log()
```

packages/cli/src/commands/inspect.ts
```diff
@@ -91,9 +98,42 @@ export async function inspect(
...
+    console.log('-------------------');
+    console.log();
+    if (sources) {
+      console.log('Contract Sources:');
+      console.log('-------------------');
+      for (let contractName in contractsAndDetails) {
+          const { sourceName, highlight } = contractsAndDetails[contractName];
+          if (sourceName) {
+              const displayName = highlight ? bold(contractName) : contractName;
+              console.log(`${displayName}: ${sourceName}`);
```
dbeal-eth, line 127: great that we have access to displaying the contract sources. This can be done in a future PR, but part of what I was thinking wrt this feature would be having the ability to dump a contract source, rather than printing out the normal output. for example, 

```
cannon inspect synthetix --source contracts/modules/AccountModule.sol
```

would simply print out the contents of that file. or the user can specify a address like:


```
cannon inspect synthetix --source AccountModule
```

would display *all* the contract sources corresponding to the account module.

packages/cli/src/helpers.ts
```diff
@@ -304,3 +306,41 @@ export function toArgs(options: { [key: string]: string | boolean | number | big
...
   });
 }
+
+/**
+ * Extracts the contract and details from the state of a deploy package
+ *
+ * @param state The deploy package state
+ * @returns an object containing ContractData
+ */
+export function getContractsAndDetails(state: { [key: string]: ArtifactContracts }): {
```
dbeal-eth, line 316: return value is kind of weird here?

packages/cli/src/helpers.ts
```diff
@@ -304,3 +306,41 @@ export function toArgs(options: { [key: string]: string | boolean | number | big
...
+
+  return contractsAndDetails
+}
+
+/**
+ *
+ * @param registries The cannon registries
+ * @returns The LocalRegistry source
+ */
+export function getSourceFromLocalRegistry(registries: CannonRegistry[]): string | undefined {
```
dbeal-eth, line 340: should always load from the registry that is configured by the user's settings. the local registry may or may not be in use by a specific user depending on circumstances.

dbeal-eth (changes requested):
the screenshots look great and the new info being displayed by the command will definitely provide fetaures to be expected by the command. as fuzzbot said, great job on the tests. we dont take great care of them here :)

however, some cases that need to be considered that would be problems in different contexts. please let me know of any questions.

PR #556, opened 2023-10-31
feat(write-scripts): add basic json dump script

mjlescano:
This PR adds the option to be able to dump the running build (regardless of it being `--dry-run` or not) into a json or ethers script file. Both files will include all the executed contract deployments, router deployments and invoke calls.

> ⚠️ To keep in mind, these scripts are stateless, this means that when re-executed they will try to use the same data generated on new, steps will not interact between them, meaning that if some step result value is dynamic and is needed for another one this will not be taken into account and it will just use the generated value from the original `cannon build` command.
> 
> So, this should be used for testing/sandboxing purposes only, as it is just a  dummy script that mimics only contracts deployments and invoke calls, and it doesn't include all the context interactions, custom signers, etc that cannon has when building from a cannonfile.
> 

Here are 2 files generated when running the following commands on synthetix's repo:
```bash
yarn hardhat cannon:build --write-script ~/Desktop/synthetix.dump.ts --write-script-format ethers-v5
yarn hardhat cannon:build --write-script ~/Desktop/synthetix.dump.json --write-script-format json
```
[synthetix.dump.json](https://github.com/usecannon/cannon/files/13221984/synthetix.dump.json)
[synthetix.dump.ts](https://github.com/usecannon/cannon/files/13221987/synthetix.dump.ts.json)

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -12,7 +12,6 @@ import {
   getContractFromPath,
   getOutputs,
 } from '@usecannon/builder';
-import { ChainBuilderContext } from '@usecannon/builder/dist/types';
 import { bold, cyanBright, gray, green, magenta, red, yellow, yellowBright } from 'chalk';
 import { ethers } from 'ethers';
 import _ from 'lodash';


dbeal-eth (changes requested):
I don't generally like the approach taken to generate here because it does not work with cannon plugins if they are being used, and even if they are not, this code is likely to become a point of maintenance vurden as either output formats change, or 

Instead, the steps should be monitored for transactions executed and those executed transactions should simply be saved in the file. This is what we do in the gnosis safe , so are we not able to do the same here?

dbeal-eth (approved):

noisekit:
1. For js export
```js
const iface = new ethers.utils.Interface(abi).format(ethers.utils.FormatTypes.full);
// 
abi: JSON.stringify(iface, null, 2)
```
abi in JS files need to be readable (for future tooling use)
2. Use JS export. Not TS. Dump files do not need extra hurdle of ts compilation and should be usable in plain nodejs with `require()`

PR #605, opened 2023-11-10
fix(cli): fix for anvil ipv6 bug

FuzzB0t:


packages/cli/src/rpc.ts
```diff
@@ -183,7 +183,7 @@ export function createProviderProxy(provider: ethers.providers.JsonRpcProvider):
 
     server.on('listening', () => {
       const addrInfo = server.address() as { address: string; family: 'IPv4' | 'IPv6'; port: number };
-      resolve(`http://${addrInfo.family === 'IPv6' ? '[' + addrInfo.address + ']' : addrInfo.address}:${addrInfo.port}`);
+      resolve(`http://127.0.0.1:${addrInfo.port}`);
```
FuzzB0t, line 186: Might need to change this back to the interpolated address so we dont restrict the ip to localhost?

Something like 
```
resolve(`http://${addrInfo.address}:${addrInfo.port}`);`
```

Unless that was what was breaking CI on Synthetixv3 in the first place but I think what broke it was the address resolution when the family was IPv6

dbeal-eth (approved):

PR #618, opened 2023-11-17
feat(hardhat-cannon): add cannon:test command

mjlescano:
This PR adds the `cannon:test` command, which basically builds the given cannonfile, adds the `hre.cannon.getContract('ContractName')` util to the context, and executes `hre hardhat test` task.


dbeal-eth (approved):
want to see some more features but this is a huge step for hardhat integration!

PR #628, opened 2023-11-22
fix(cli): prevent name/version remote registry match on build

StuartGavidia:
Prevent name/version remote registry match on build
Create --skip-reference to skip the check

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -52,7 +52,7 @@ interface Params {
   priorityGasFee?: string;
   writeScript?: string;
   writeScriptFormat?: WriteScriptFormat;
-  skipReference: boolean;
+  ignorePkgrefCheck: boolean;
 }
 
 export async function build({
@@ -77,7 +77,7 @@ export async function build({
   priorityGasFee,
   writeScript,
   writeScriptFormat = 'ethers',
-  skipReference = false,
+  ignorePkgrefCheck = false,
 }: Params) {
   if (wipe && upgradeFrom) {
     throw new Error('wipe and upgradeFrom are mutually exclusive. Please specify one or the other');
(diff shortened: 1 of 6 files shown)


FuzzB0t:
Seems like some tests are failing but that might just be because you need to merge main into your branch.

dbeal-eth (changes requested):
appears that this change needs to be rebased? please reach out to me if you need support with this

packages/cli/src/commandsConfig.ts
```diff
@@ -293,6 +293,10 @@ const commandsConfig = {
         description: '(Experimental) Format in which to write the actions script (Options: json, ethers)',
         defaultValue: 'ethers',
       },
+      {
+        flags: '-sr --skip-reference',
```
dbeal-eth, line 297: the naming of this option does little to explain what it does (`reference` is an extremely broad concept)

perhaps a name such as `--ignore-pkgref-check` would ? there isnt a lot of pressure for it to be short because we dont want people to use it unless they know what they are doing also :sweat_smile:

packages/cli/src/commandsConfig.ts
```diff
@@ -293,6 +293,10 @@ const commandsConfig = {
         description: '(Experimental) Format in which to write the actions script (Options: json, ethers)',
         defaultValue: 'ethers',
       },
+      {
+        flags: '-igc --ignore-pkgref-check',
```
dbeal-eth, line 297: a single dash flag cn only have one character. ex `-i`

packages/cli/src/commands/build.ts
```diff
@@ -148,6 +150,13 @@ export async function build({
 
   const resolver = overrideResolver || (await createDefaultReadRegistry(cliSettings));
 
+  const uri = await resolver.getUrl(fullPackageRef, chainId);
```
dbeal-eth, line 153: 2 possibly major issues here:
1. we only want to check that the package does not exist on the `OnChainRegistry`. It is not a problem if the package exists/is resolvable on the local registry (in fact, that is encouraged!)
2. we should not call `await resolver.getUrl` if `ignorePkgrefCheck` is set to true, since it is a completely wasted call

dbeal-eth (changes requested):

dbeal-eth:
@StuartGavidia please fix conflicts too when you get a chance!

PR #643, opened 2023-11-29
fix(cli): Interact ref bug

FuzzB0t:
Related to this:

image


FuzzB0t:
Decided to split this change off because the interact is completely broken in the current release

dbeal-eth (approved):

PR #645, opened 2023-11-30
fix: Preset regressions

FuzzB0t:
fixes https://linear.app/usecannon/issue/CAN-140


packages/builder/src/definition.ts
```diff
@@ -292,11 +293,26 @@ export class ChainDefinition {
     // it would be best if the dep was downloaded when it was discovered to be needed, but there is not a lot we
     // can do about this right now
     return _.uniq(
-      Object.values(this.raw.import).map((d) => ({
-        source: _.template(d.source)(ctx),
-        chainId: d.chainId || ctx.chainId,
-        preset: _.template(d.preset || 'main')(ctx),
-      }))
+      Object.values(this.raw.import).map((d) => {
```
FuzzB0t, line 296: Not sure these changes are even relevant or correct, the getRequiredImports function doesnt even get called anywhere else.

packages/builder/src/registry.ts
```diff
@@ -27,25 +27,25 @@ export abstract class CannonRegistry {
...
-      return result.indexOf('@', 1) !== -1 ? result.slice(0, result.indexOf('@', 1)) : result;
+    if (serviceRef.startsWith('@')) {
+      const result = serviceRef.replace(':', '://').replace('@', '');
+      return result.indexOf('@') !== -1 ? result.slice(0, result.indexOf('@')) : result;
     }
 
     return null;
   }
 
+  // TODO: why is this empty?
```
dbeal-eth, line 40: It's empty because there's never a case where we would want to return data in this superclass of the Canon registry whose purpose is to simply be a base common implementation

Perhaps we can make it abstract by having the function be a headless definition?

dbeal-eth (approved):

PR #649, opened 2023-11-30
fix(cli): Implement global debug level settings using Commander lifecycle hooks

StuartGavidia:
Added debugging flags to following cannon commands:
* run
* verify
* alter
* fetch
* pin
* publish
* inspect
* prune
* trace
* decode
* test
* interact
* setup
* clean
* plugin list
* plugin add
* plugin remove


dbeal-eth (approved):
just to confirm, this has been tested? love it.

PR #675, opened 2023-12-06
fix: Fixes on publish with preset

FuzzB0t:


packages/builder/src/package.ts
```diff
@@ -176,14 +192,14 @@ export async function publishPackage({
...
-  preset,
   includeProvisioned = false,
 }: CopyPackageOpts) {
   debug(`copy package ${packageRef} (${fromStorage.registry.getLabel()} -> ${toStorage.registry.getLabel()})`);
 
   // TODO: packageRef in this case can be a package name or an IPFS hash (@ipfs://Qm...) for the pin command, however, this functionality should have
   // it's own function to handle the pinning of IPFS urls.
-  const fullPackageRef = !packageRef.startsWith('@') ? new PackageReference(packageRef).fullPackageRef : packageRef;
+  const presetRef = PackageReference.isValid(packageRef) ? new PackageReference(packageRef).preset : 'main';
+  const fullPackageRef = PackageReference.isValid(packageRef) ? new PackageReference(packageRef).fullPackageRef : packageRef;
```
dbeal-eth, line 202: redundant

packages/builder/src/registry.ts
```diff
@@ -37,6 +37,12 @@ export abstract class CannonRegistry {
 
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   async getMetaUrl(serviceRef: string, chainId: number): Promise<string | null> {
+    // Check if its an ipfs hash / url, if so we make sure to remove any incorrectly appended presets (like @main);
+    if (serviceRef.startsWith('@')) {
+      const result = serviceRef.replace(':', '://').replace('@', '');
```
dbeal-eth, line 42: code here is exactly the same as the `getUrl` function.

I do not think this change should be implemented, reasoning below.

if it is determined that the meta url should match the plain url for a (ex) `@ipfs:Qmarst`, then its a change of functonality and we probably need to do some more testing

however, I would not reccomend this change because previously the funtionality was just to return null because, if you think about it, if all you have is just an IPFS hash, then there isnt a way to resolve the metadata corresponding to this ipfs hash. to attempt to do so could result in bugs/crashing when apps expect metadata to come back in one format, and it turns out to be another.

packages/builder/src/steps/provision.test.ts
```diff
@@ -84,7 +84,7 @@ describe('steps/provision.ts', () => {
       const result = await action.getState(
         fakeRuntime,
         fakeCtx,
-        { source: 'hello:1.0.0', sourcePreset: 'foobar', chainId: 1234, targetPreset: 'voop', options: { bar: 'baz' } },
```
dbeal-eth, line 87: tests were previously testing with foobar rather than main likely for a reason... why did it become necessary to change to main?

packages/cli/src/registry.ts
```diff
@@ -88,16 +88,17 @@ export class LocalRegistry extends CannonRegistry {
     return [];
   }
 
+  //
```
dbeal-eth, line 91: empty comment

packages/builder/src/steps/provision.test.ts
FuzzB0t: The package references in the tests are using main as the preset and the tests were failing because it was prioritizing the sourcePreset.

PR #681, opened 2023-12-11
fix(builder): Cannon build fixes

FuzzB0t:
Still a work in progress, so far:

- Added a better error output for invalid dependencies
- Encode contract call args before executing to check for any argument type errors and present them to the user, Example:

Before when using an incorrect argument (in this example a non-address used in an address field) this was the error:
image

Now it looks more like this: 
image

Fixes https://linear.app/usecannon/issue/CAN-137/
Fixes https://linear.app/usecannon/issue/CAN-101/
Fixes https://linear.app/usecannon/issue/CAN-108/


packages/builder/src/steps/invoke.test.ts
```diff
@@ -202,8 +202,11 @@ describe('steps/invoke.ts', () => {
         {
           target: ['Woot'],
           func: 'something',
-          args: ['foo', { bar: 'baz' }],
-          value: '1234',
+          args: [
```
dbeal-eth, line 205: why were these args changed? having a nested argument/tuple arg is part of the test

packages/builder/src/steps/invoke.ts
```diff
@@ -95,6 +95,13 @@ async function runTxn(
     overrides.maxPriorityFeePerGas = runtime.priorityGasFee;
   }
 
+  // Attempt to encode data so that if any arguments have any type mismatches, we can catch them and present them to the user.
+  try {
+    contract.interface.encodeFunctionData(config.func, config.args);
+  } catch (error: any) {
+    throw new Error(`Invalid arguments for function "${config.func}": \n\n ${error}`);
```
dbeal-eth, line 102: we may be able to inspect the data structure of `error` to see if it includes any additional useful data we can pretty print

packages/cli/src/registry.ts
```diff
@@ -174,18 +181,24 @@ export async function createDefaultReadRegistry(
...
   } else if (settings.registryPriority === 'local') {
     debug('local registry is the priority, using local registry first');
-    return new FallbackRegistry([...additionalRegistries, localRegistry, onChainRegistry]);
+    return new FallbackRegistry([...additionalRegistries, inMemoryRegistry, localRegistry, onChainRegistry]);
   } else {
     debug('on-chain registry is the priority, using on-chain registry first');
-    const fallbackRegistry = new FallbackRegistry([...additionalRegistries, onChainRegistry, localRegistry]);
+    const fallbackRegistry = new FallbackRegistry([
+      ...additionalRegistries,
+      inMemoryRegistry,
```
dbeal-eth, line 198: in memory registry should always be first in the array, not after additionals.

dbeal-eth: also, please *double check* the `FallbackRegistry` as it is now, I thik it has a `InMemoryRegistry` built into it now, so its unnecessary (and harmful!) to add the in memory registry here.

packages/cli/src/rpc.ts
```diff
@@ -183,7 +183,9 @@ export function createProviderProxy(provider: ethers.providers.JsonRpcProvider):
     });
 
     server.on('listening', () => {
-      const addrInfo = server.address() as { address: string; family: 'IPv4' | 'IPv6'; port: number };
+      const address = server.address();
+      const addrInfo = address as { address: string; family: 'IPv4' | 'IPv6'; port: number };
```
dbeal-eth, line 187: why separate the retyping into a separate var?

packages/hardhat-cannon/src/tasks/build.ts
```diff
@@ -100,24 +102,26 @@ task(TASK_BUILD, 'Assemble a defined chain and save it to to a state which can b
...
-          ? await runRpc(
-              {
-                port: hre.config.networks.cannon.port || anvilOpts.port,
-                chainId: (await hre.ethers.provider.getNetwork()).chainId,
-                accounts: anvilOpts.accounts || 10,
-                ...anvilOpts,
-              },
-              {
-                forkProvider: new ethers.providers.JsonRpcProvider(providerUrl),
-              }
```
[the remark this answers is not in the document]
dbeal-eth: :+1:

dbeal-eth (changes requested):
please confirm a few things

packages/hardhat-cannon/src/tasks/build.ts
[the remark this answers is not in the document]
FuzzB0t: Running into this error when adding the forkProvider:

image

Im guessing it has something to do with the port mismatch between the default port and the forkUrl port of the anvil instance, however when running the anvil instance with the specified port anvil simply times out.

packages/builder/src/steps/invoke.test.ts
FuzzB0t: Changed them back, I did need to however add another argument as this was causing the test to fail saying it requires 3 arguments because of the argument encode check I added in this branch:

image

packages/cli/src/registry.ts
FuzzB0t: FallbackRegistry did have the InMemoryRegistry built into it, so I removed the changes I added.

packages/cli/src/rpc.ts
FuzzB0t: Was using the var to console.log and seemingly did not remove it.

packages/hardhat-cannon/src/tasks/build.ts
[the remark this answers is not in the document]
dbeal-eth: any progress on this? the fact that its having an error does not justify removing it. if you are stuck, please let me know

dbeal-eth (changes requested):
need to fix lints. also see comment

packages/hardhat-cannon/src/tasks/build.ts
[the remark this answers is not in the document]
FuzzB0t: @dbeal-eth Tackling this issue today alongside @saeta-eth, we're gonna pair up on it and get this ticket over the line ASAP.

packages/hardhat-cannon/src/tasks/build.ts
```diff
@@ -100,24 +102,29 @@ task(TASK_BUILD, 'Assemble a defined chain and save it to to a state which can b
...
+
       if (dryRun || hre.network.name === 'cannon') {
+        const port = anvilOpts.port || hre.config.networks.cannon.port;
+        const chainId = anvilOpts.chainId || 13370;
+        const accounts = anvilOpts.accounts || Number.parseInt(hre.config.networks.cannon.accounts.toString()) || 10;
         const node = dryRun
           ? await runRpc(
               {
-                port: hre.config.networks.cannon.port || anvilOpts.port,
-                chainId: (await hre.ethers.provider.getNetwork()).chainId,
```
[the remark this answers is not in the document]
FuzzB0t: Adding this line is causing the following error not entirely sure why: 
image

PR #686, opened 2023-12-14
fix(hardhat-cannon): Updates for compatibility with ethers v6

mjlescano:


dbeal-eth (approved):
i know this isnt commont right now, but what happens if people use hardhat-viem?

its not the oslution I was expecting, but its ok for now

dbeal-eth:
also please fix the lints and etc

dbeal-eth:
@mjlescano is this ready for merge/review? why is it drafted?

PR #711, opened 2024-01-10
fix(cli): Get evm version directly from `forge config --json`

saeta-eth:
This PR aims to fix an error related to MacOS lacking native support for `jq`, which requires manual installation.

```
Error: Failure on step contract.EmptyContract: Error: Command failed: forge config --json | jq .evm_version
/bin/sh: jq: command not found
The application panicked (crashed).
Message:  failed printing to stdout: Broken pipe (os error 32)
Location: library/std/src/io/stdio.rs:1019

This is a bug. Consider reporting it at https://github.com/foundry-rs/foundry

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

### DIFF
--- a/packages/website/src/features/Settings/SettingsPage.tsx
+++ b/packages/website/src/features/Settings/SettingsPage.tsx
@@ -131,7 +131,10 @@ export default function SettingsPage() {
               </Link>{' '}
               endpoints) added below if available for the target chain.
               Otherwise, it will use a{' '}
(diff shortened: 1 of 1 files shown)


dbeal-eth (approved):
ah woops. yea definitely need to change this. dont know why I had jq set up here :rofl:

PR #725, opened 2024-01-19
fix(cli): update cannon clean to clear all cache files

StuartGavidia:


dbeal-eth (approved):
thanks for identifying this and fixing!

PR #744, opened 2024-01-30
fix(website): remove ethers for viem

mjlescano:


packages/website/src/features/Deploy/PublishUtility.tsx
```diff
@@ -72,15 +76,14 @@ export default function PublishUtility(props: {
       );
 
       const targetRegistry = new OnChainRegistry({
```
dbeal-eth, line 78: its probably also necessary to supply `provider` here along with the signer because of the way viem works

packages/website/src/features/Deploy/PublishUtility.tsx
```diff
@@ -104,7 +107,6 @@ export default function PublishUtility(props: {
 
       await publishPackage({
         packageRef: `${resolvedName}:${resolvedVersion}@${resolvedPreset}`,
-        // TODO: Check if we need to provide tags
```
dbeal-eth, line 107: technically I think this is still a todo?

packages/website/src/features/Packages/Function.tsx
```diff
@@ -132,10 +124,13 @@ export const Function: FC<{
         // console.error(e);
         // setError(e?.message || e?.error?.message || e?.error || e);
         try {
+          /*
```
dbeal-eth, line 127: shoudl we just get rid of this code if we have established its not needed? if it was needed for osmething, should we get this filled in?

packages/website/src/helpers/onchain-store.ts
```diff
@@ -86,9 +84,9 @@ export const ABI = [
   },
 ];
 
-export const deployTxn: Partial<TransactionRequestBase> = {
-  to: '0x4e59b44847b379578588920ca78fbf26c0b4956c',
-  data: (hexZeroPad('0x', 32) + bytecode) as Hex,
+export const deployTxn = {
+  to: '0x4e59b44847b379578588920ca78fbf26c0b4956c' as viem.Hex,
+  data: (viem.pad('0x', { size: 32 }) + bytecode) as viem.Hex,
```
dbeal-eth, line 89: does this code produce the same output result?

packages/website/src/hooks/backend.ts
```diff
@@ -130,21 +120,25 @@ export function useTxnStager(
...
-
-    const signers = [];
-    for (const sig of alreadyStaged.sigs) {
-      const regularSig = ethers.utils.arrayify(sig);
-      regularSig[regularSig.length - 1] -= 4;
-      signers.push(ethers.utils.verifyMessage(ethers.utils.arrayify(hashToSign), regularSig));
-    }
+  const hashToSign = reads.isSuccess ? (reads.data![0].result as viem.Address) : null;
+
+  useEffect(() => {
```
dbeal-eth, line 125: its really a shame that viem requires `await` for recoverAddress. When I checked the source code, the reason is because it dynamically `import`s a module from inside the recover function. Not sure why they couldn't just let people import this module if they want it.

dbeal-eth (approved):
this turned out to be more a general purpose refactor/code cleanup, not just to the website.

not that I am complaining :+1: high quality code here I think

btw, does the `safeService` library that we are using use ethers or anything like that? if so, we may still have a hidden ethers dependency

PR #753, opened 2024-02-05
fix: enable console logs lint and others

dbeal-eth:


dbeal-eth:
fixes https://linear.app/usecannon/issue/CAN-180/add-lint-rule-for-consollog-disallowed-in-most-places
fixes https://linear.app/usecannon/issue/CAN-179/check-build-package-outputs

FuzzB0t (approved):

PR #762, opened 2024-02-13
fix(cli): fix getUrl function

saeta-eth:
fixes https://linear.app/usecannon/issue/CAN-188


FuzzB0t (approved):

FuzzB0t:
fixes https://linear.app/usecannon/issue/CAN-188

PR #772, opened 2024-02-16
fix(website): fix anchor links on interact section

saeta-eth:


noahlitvin (changes requested):
Awesome! Is it possible to fix to adjust for the height of the sticky header on desktop? e.g. it anchors correctly, but then I need to scroll up to reveal the target.

noahlitvin (approved):

PR #777, opened 2024-02-20
feat: add cannon indexer

dbeal-eth:
this can be used to populate a redis database with indexed data from IPFS and on-chain sources of cannon data. the kinds of data that is indexed includes:

* address to package
* transaction to package
* selectors lookup and matching to contracts that use them
* packages depending on packages
* number of times steps are used
* various time series records

and others

also includes some minor changes to the builder to improve loading of packages


noahlitvin:
LGTM. Tests are passing for me locally after rebuilding the builder.

noahlitvin (approved):

FuzzB0t (approved):

PR #790, opened 2024-02-24
fix: router deployment with viem

mjlescano:


dbeal-eth (approved):
Still don't see simulate transaction anywhere, but if you think it's not necessary then I'm okay with not having it I suppose

PR #791, opened 2024-02-25
feat(builder): Cannonfile Spec Renaming

dbeal-eth:
This pull request includes a variety of changes to the [Cannonfile specification](https://usecannon.com/learn/cannonfile) that make the developer experience more intuitive when writing Cannonfiles. Existing naming conventions will be deprecated but continue to work, so this will be a non-breaking change.

**We are very interested in community feedback on everything here—especially the step/variable naming—proposed below.**

### Rename `setting` to `var` + combine `settings` and `extras` into `vars`

Currently, we have the ability to define settings in a Cannonfile, which can be overridden when running a build/deployment but never change during the build/deployment. We also have the ability to store data emitted from events under the name `extras`. (Originally, the intention was more than just event data could be captured under this namespace, but no other use cases emerged.)

This combines these two types of data into `vars` (short for *variables*). It also deprecates `setting` in favor of `var`, which a proper step that can update `vars` based on data in other steps. These vars are all available available in the resulting cannon package, which can be imported/pulled into other cannonfiles and inspected using the web explorer/CLI.

### Rename `contract` step to `deploy`

The contract step is used to deploy a contract. Having a step defined as `[contract.my_example_contract]` and then having the resulting data referenced in later steps in the plural (`{% contracts.my_example_contract %}`) has led to confusion. Also, the `router` step and `invoke` factory functionality both write to the `contracts` namespace, leading to confusion as well.

### Rename `import` step to `pull`

Similar to the above change, the `import` step writes data under `imports` to be referenced in later steps. `imports` data is also written to by the `provision` step. This name (which means to suggest "pulling in package data") should reduce this confusion.

### Remaining Tasks
* Update all the docs
* Add some tests
* Make a migration script for renaming cannonfiles and updating state files.


FuzzB0t (approved):

noahlitvin:
I was also thinking about `[variable.x]` referenced with `variables.x`.

noahlitvin:
I like `set` for the step because we could keep them referenced as `settings.x` 

Settings files would look like

```
[set.eth_perps_market]
funding_rate_velocity = 420
maximum_open_interest = 69
```

Honestly the implementation like this instead of with `defaultValue` makes me less concerned with setting/settings. `extras` could get renamed to `settings`.

examples/sample-foundry-project/cannonfile.consumer.toml
```diff
@@ -38,19 +38,19 @@ target = ["greeters.greeter"]
...
+var.OldCloneGreeting.event = "OldGreetingRemoved"
+var.OldCloneGreeting.arg = 0
+var.OldCloneGreeting.artifact = "Greeter"
 
 # test to parse through previous emitted event values
 [invoke.set_new_greeting_for_next_clones]
 target = ["greeters.greeter"]
 func = "setGreeting"
-args = ["<%= extras.NewCloneGreeting %>"]
+args = ["<%= settings.NewCloneGreeting %>"]
```
noisekit, line 53:
```suggestion
args = ["<%= var.NewCloneGreeting %>"]
```

examples/sample-foundry-project/cannonfile.consumer.toml
```diff
@@ -38,19 +38,19 @@ target = ["greeters.greeter"]
...
-extra.NewCloneGreeting.arg = 0
-extra.NewCloneGreeting.artifact = "Greeter"
+var.NewCloneGreeting.event = "NewGreetingAdded"
+var.NewCloneGreeting.arg = 0
+var.NewCloneGreeting.artifact = "Greeter"
 
-extra.OldCloneGreeting.event = "OldGreetingRemoved"
-extra.OldCloneGreeting.arg = 0
-extra.OldCloneGreeting.artifact = "Greeter"
+var.OldCloneGreeting.event = "OldGreetingRemoved"
```
noisekit, line 45:
```suggestion
var.OldCloneGreeting = "events.OldGreetingRemoved[0]"
```

examples/sample-foundry-project/cannonfile.consumer.toml
noisekit: It is very important to match all instances of the full string because we are not dealing with programming language but with strings in toml and can only rely on full text search when performing refactoring.

examples/sample-foundry-project/cannonfile.consumer.toml
```diff
@@ -3,18 +3,18 @@ version = "0.0.1"
 description = "Simple project to verify the functionality of cannon"
 keywords = ["sample", "consumer"]
 
-[setting.change_greeting2]
-defaultValue = "new greeting!"
+[var.change_greeting2]
+value = "new greeting!"
 
-[import.greeters]
+[pull.greeters]
```
noisekit, line 9: `import` is more suitable. There is no pull happening at this step, we import deployed contracts (addresses) from chain.

examples/sample-foundry-project/cannonfile.consumer.toml
```diff
@@ -3,18 +3,18 @@ version = "0.0.1"
...
 source = "greeter-foundry"
 
-[import.more_greeters]
+[provision.more_greeters]
```
noisekit, line 12:
```suggestion
[deploy.more_greeters]
```

examples/sample-foundry-project/cannonfile.consumer.toml
noahlitvin: I generally agree with this, but I'm sympathetic to the point that `import` in the step and `imports` in the context trips people up.

examples/sample-foundry-project/cannonfile.consumer.toml
noahlitvin: I like this. The other consideration is (like the other point) differentiating the step name from how it's referenced in the context. Are you imagining 'contracts' stays the same?

noisekit: Yeah I recon contract is ok.
Tbh it does not matter that much how things  are called, what really does matter is that name matches when you declare and when you use a thing. 
If I call something contract.Greeter, I expect to reference it everywhere in the code as contract.Greeter and not anything else.

dbeal-eth:
> 
> ```
> [settings.eth_perps]
> eth_perps_funding_rate_velocity = 420
> eth_perps_maximum_open_interest = 69
> 
> ```
> 
> You reference `settings. eth_perps_maximum_open_interest`
> 
> `extras` could get renamed to `settings` in `invoke`
> 
> Maybe namespace it in the context with the step name?

I really like the idea of setting settings directly in groups like your example here, but I don't like calling the step name `settings`..., `variable` or `set` would be better

examples/sample-foundry-project/cannonfile.consumer.toml
noahlitvin: Got it, so you'd just expect to reference with `contract` and `import` (singular)?

noisekit: Yes 100%

examples/sample-foundry-project/cannonfile.consumer.toml
noisekit: May I suggest to not rename anything that already exists and just focus on what was giving the most pain. After some extra reading, new names do not make more sense than old ones, but introducing such a big breaking change will add a lot of friction.

PR #814, opened 2024-03-09
fix(hardhat): provider initialization

mjlescano:


dbeal-eth (approved):
tested on synthetix so seems good.

would be good to document the syptoms/occasions on which this problem occurs, since its not clear just looking at the fix

PR #817, opened 2024-03-11
fix(builder): reduced access interpolation notation

StuartGavidia:
Allowing users to use a simplified syntax to access contracts in interpolation on cannon invoke steps. 

Succesfully tested that these steps built in a cannonfile:
[invoke.checkIfContract]
target = ["greeters.greeter"]
func = "checkIfContract"
args = ["<%= imports.greeters.contracts.greeter.address %>"]

[invoke.checkIfContract]
target = ["greeters.greeter"]
func = "checkIfContract"
args = ["<%= greeters.greeter.address %>"]


packages/builder/src/types.ts
```diff
@@ -83,6 +83,8 @@ export interface ChainBuilderContext extends PreChainBuilderContext {
   extras: { [label: string]: string };
 
   imports: BundledChainBuilderOutputs;
+
+  [additionalProperties: string]: any;
```
dbeal-eth, line 87: call this perhaps `shortContract`? and you may be able to define the type more precisely here than `any` for the value!

packages/builder/src/builder.ts
```diff
@@ -385,19 +385,31 @@ export async function getOutputs(
 // TODO: this func is dumb but I need to walk through this time period before I want to turn it into something of beauty
 function addOutputsToContext(ctx: ChainBuilderContext, outputs: ChainArtifacts) {
   const imports = outputs.imports;
-
   for (const imp in imports) {
     ctx.imports[imp] = imports[imp];
+    //also add simplified syntax
+    ctx[imp] = { ...imports[imp], contracts: undefined };
+    delete ctx[imp].contracts;
```
dbeal-eth, line 392: does this remove the "complicated" syntax? for backwards compatibility we want to keep this syntax alive for the time being.

packages/builder/src/builder.ts
```diff
@@ -385,19 +385,31 @@ export async function getOutputs(
...
+
+    if (imports[imp].contracts != undefined) {
```
dbeal-eth, line 394: `!= undefined` is usually just `!imports[imp].contracts` in this codebase

dbeal-eth (changes requested):
great work!

does this code still work if you have an import inside of an import?

ex. `imports.subpkg.imports.subsubpkg.contracts.MyContract` should also work (and more and more layers deeper). So you should be recursing.

please add some tests to the `addOutputsToContext` function to verify correct functionality here.

packages/builder/src/builder.ts
StuartGavidia: will fix!

dbeal-eth (approved):
great work on this!

dbeal-eth:
@StuartGavidia sorry to ask more of you, but could you please investigate why `test-e2e` is failing? if it appears to be unrelated to your PR, you may merge.

PR #827, opened 2024-03-19
chore(registry): fix styles and lint errors

mjlescano:


packages/registry/contracts/CannonRegistry.sol
```diff
@@ -293,19 +276,26 @@ contract CannonRegistry is EfficientStorage, OwnedUpgradable {
...
+  }
+
   function _checkCrossDomainSender() internal {
     // we can only receive change ownership requests from our counterpart on mainnnet
     if (_OPTIMISM_RECEIVER.xDomainMessageSender() != address(this)) {
       revert Unauthorized();
     }
   }
 
-  function _propogatePackageOwner(bytes32 _packageName, address _owner, address[] memory _additionalDeployers) internal {
```
FuzzB0t, line 303: Why did we remove this?
We just deployed a version of the CannonRegistry so this will break the partial deployment i believe

PR #828, opened 2024-03-20
fix: misc fixes from name respec

FuzzB0t:


dbeal-eth:
how lcose is this to merge @FuzzB0t ?

FuzzB0t:
Still seeing an error on publish reverting with "Unauthorized()", this is the last piece before the e2e tests are working again. 
SNX ci still seeing some issues but not sure whats causing them.

packages/cli/src/commands/alter.ts
```diff
@@ -166,13 +167,48 @@ export async function alter(
...
-        deployInfo.state[stepName].artifacts = await stepAction.importExisting(
+
+        const importExisting = await stepAction.importExisting(
           runtime,
           ctx,
           config,
           { currentLabel: stepName, name: def.getName(ctx), version: def.getVersion(ctx) },
           existingKeys
         );
+
```
FuzzB0t, line 178: These changes are required for when a partial deployment has an empty step, when running `alter import` this would cause the command to fail.

The idea is it autopopulates the step with an empty value and then importExisting does the rest of the work.

packages/builder/src/registry.ts
```diff
@@ -274,7 +274,26 @@ export class OnChainRegistry extends CannonRegistry {
     }
   }
 
-  private async checkPackageOwnership(packageName: string) {
+  /**
```
FuzzB0t, line 277: Adding a check for when a new package is being published and needs to be 'registered'. 
Subsequently register the package if the check doesnt pass.

packages/builder/src/definition.ts
```diff
@@ -83,7 +83,12 @@ export class ChainDefinition {
         }
 
         const fullActionName = `${action}.${name}`;
-        actionNames.push(name);
+
```
FuzzB0t, line 86: @dbeal-eth Im a little curious about this change and would appreciate your input.

Before on one of the snx builds we were getting a duplicate error due to a setting taking up the name, now that setting is a step I figure we dont want to do this but this was the only way I could restore backwards compatibility without breaking the rest of the 'var' action registering.

packages/builder/src/registry.ts
```diff
@@ -308,19 +327,36 @@ export class OnChainRegistry extends CannonRegistry {
...
+          data.metaUrl || '',
+        ],
+      };
+
+      // if package is not registered, we register it (aka set ownership) here.
+      if (!(await this.isPackageRegistered(data.name))) {
+        const registerTx = {
+          abi: this.contract.abi,
+          address: this.contract.address,
+          functionName: 'setPackageOwnership',
```
dbeal-eth, line 351: This does not work on optimism

packages/builder/src/steps/pull.ts
```diff
@@ -52,7 +52,7 @@ const pullSpec = {
     const packageRef = new PackageReference(_.template(config.source)(ctx));
 
     config.source = packageRef.fullPackageRef;
-    config.preset = _.template(config.preset)(ctx);
+    config.preset = _.template(config.preset)(ctx) || packageRef.preset;
```
dbeal-eth, line 55: Why was this adder here? Should it go somewhere else? How did it work in prev versions

dbeal-eth (approved):

packages/builder/src/definition.ts
dbeal-eth: Not sure what your asking here, need to look closer over weekend perhaps

dbeal-eth: Not sure how they can have duplicate names. Could you share example?

FuzzB0t: In synthetix-sandbox we have this setting here:

https://github.com/Synthetixio/synthetix-sandbox/blob/main/cannonfile.toml#L11

and then there is a provision step by the same name:

https://github.com/Synthetixio/synthetix-sandbox/blob/main/cannonfile.toml#L37

PR #834, opened 2024-03-22
fix(cli): run command supports private key

saeta-eth:
This PR aims to fix an issue on the run command but also on the build command when it's executed with `--keep-alive`. Both commands were using `0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266` ignoring the account behind the private key.

Command examples to reproduce the issue and check the fix:

> `cannon package-name --provider-url <RPC> --private-key <PK>`

> `cannon build something.toml --provider-url <RPC> --private-key <PK> --dry-run --keep-alive`

Closes https://linear.app/usecannon/issue/CAN-247/run-command-should-support-private-key

### DIFF
--- a/packages/cli/src/commands/run.ts
+++ b/packages/cli/src/commands/run.ts
@@ -25,6 +25,7 @@ import onKeypress from '../util/on-keypress';
 import { build } from './build';
 import { interact } from './interact';
 import { TraceEntry } from '@usecannon/builder/src';
+import { normalizePrivateKey } from '../util/provider';
 
 export interface RunOptions {
   node: CannonRpcNode;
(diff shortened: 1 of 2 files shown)


dbeal-eth (changes requested):
missing `0x` prefix in private key (as is often the way its exported/shared) does not work:

```
xcannon synthetix --provider-url https://ethereum-rpc.publicnode.com --private-key ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
```

(above private key is for the testnet isgner so no I did not dox my key)

`provider.ts` has a `normalizePrivatekey` function. Lets make sure we are using this *any time* a private key is used as a parameter for anything.

please add e2e test to verify the new functionality, and please also if you can, find a way to ensure that private key is *always* normalized so we can avoid this mistake in the future.

PR #839, opened 2024-03-25
fix(builder): handling of errors when deploying contracts (create2)

saeta-eth:
This PR aims to show the error when a contract reverts during deployment. It will handle deployments with CREATE2.

Closes https://linear.app/usecannon/issue/CAN-204/verify-contract-create-transaction-before-sending-to-network


dbeal-eth (changes requested):
do we have a test verifying that this decoding functionality works?

is it possible that, rather than utilizing low level error decoding capabilities, we detect *any error that cannon has awareness of, even in other contracts, so far?*

ideally we use the `CannonTraceError`, which prints out error and stack trace using data from all contracts in the project

The problem right now is that the "error" may be unknown if the constructor is executed and we havent imported the contract ABI yet.

FuzzB0t (approved):

dbeal-eth (changes requested):
on a real network, wouldn't this actually run the deployment of said contract?

a better approach to solving this problem would be to run the txn as we are now, and afterwards, check to ensure that code has been written to the address using `provider.getBytecode`. If there is no code (ie it is "null"), then that means the deployment actually failed.

dbeal-eth (approved):

PR #840, opened 2024-03-25
fix(simulation): add polling interval for simulations

mjlescano:
So, when running simulations, we execute a tx and wait for it, but the `waitForTransactionReceipt` had a default of `4_000ms` (not sure were it comes from): https://viem.sh/docs/actions/public/waitForTransactionReceipt#pollinginterval-optional

Setting that value to something sane appears to fix long waits.


dbeal-eth:
Does reducing this value in other parts of the codebase speed up simulations? Perhaps we should do this every time we build with anvil.

dbeal-eth (approved):

PR #844, opened 2024-03-27
feat: ipfs add to mfs

avniculae:


packages/builder/src/ipfs.ts
```diff
@@ -145,7 +145,7 @@ export async function writeIpfs(
 
   let result: AxiosResponse<any, any>;
   try {
-    result = await axios.post(ipfsUrl.replace('+ipfs', '') + '/api/v0/add?local=true', formData, {
```
dbeal-eth, line 148: Unfortunately we cannot remove this line; it causes breakage on `ipfs-cluster` servers in misc circumstances. Does the request still work with infura with both query string options?

packages/builder/src/ipfs.ts
```diff
@@ -145,7 +145,7 @@ export async function writeIpfs(
...
+    result = await axios.post(ipfsUrl.replace('+ipfs', '') + '/api/v0/add?to-files=%2F' + cid, formData, {
```
dbeal-eth, line 148: Would love an explanation on what this does and why it works

dbeal-eth (changes requested):
Thanks for your contribution! When we first built cannon, Infua IPFS was the first IPFS service we endorsed. However, we soon discovered a major issue with it that has lead us to not supporting it officially: The ipfs files uploaded to Infura IPFS, while resolvable within Infura endpoint, are not resolvable through any other general ipfs node. This means Infura is basically S3 with an IPFS API. If you can test/verify that this is no longer the case that would be helpful!

dbeal-eth:
Also , we are aware of the `offline` tag is preferred over `local`, however, the behavior of the offline tag is different from local, and it doesn't give us the functionality we need. So as of last time we tested all of this, `local` flag seems to be our only option

packages/builder/src/ipfs.ts
avniculae: [8c31779](https://github.com/usecannon/cannon/pull/844/commits/8c317791550e0ca0e5d068edf21d48fde8f84458)

dbeal-eth (approved):

PR #867, opened 2024-04-03
fix(builder): always rerun `clone` to ensure update

noahlitvin:
consider that this will update the timestamp, which effects the resulting Qm hash


dbeal-eth:
before merging this PR we need to add capability to prevent bumping timestamp/any other changes to package when there are no underlying changes

dbeal-eth:
test is failing due to known reason, fixed in other Pr

PR #868, opened 2024-04-04
feat(registry): add `register` command for creating packages

mjlescano:
This PR adds the new `cannon register <packageName>` command to register for the first time on the main registry a new package. This is the first for unlocking new package publications.

**Next steps (on a separate PR):**
1. Merge functionality on `publish` command. This gets tricky as the publish command will have to check if the package is not registered, register it. If not, try to infer on which registry it should be published.
2. Add possibility to migrate packages from main registry to secondary one

**Extra**
* This PR also includes some heavy refactors on builders `registry.ts` file, but functionality is exactly the same.

### DIFF
--- a/.github/workflows/test-e2e.yml
+++ b/.github/workflows/test-e2e.yml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 jobs:
-  test:
+  test-e2e:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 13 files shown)


dbeal-eth (approved):
lot of random changes seem to be all over the place, makes it a bit harder to evaluate this PR.

overall looks OK as long as you have tested everything tho :+1:

FuzzB0t (approved):

PR #873, opened 2024-04-06
feat(website): tuples as json (read-only) and render each input for tuples in write mode

Ignusmart:
The primary goal of this pull request is to enable the rendering of tuples as JSON when they are in read-only mode, and to facilitate the rendering of individual inputs for tuples in write mode.

# Deploy View (Read Only)

## Before the adjustment

image

## With the adjustment

image

## Before the adjustment

image

## With the adjustment

image

# NOTES

Before merging, we should thoroughly test the new tuple functionality, particularly in write mode, especially in scenarios where functions expect nested tuples.


packages/website/src/features/Packages/FunctionInput.tsx
```diff
@@ -31,10 +31,6 @@ export const FunctionInput: FC<{
     setDataArray(dataArray.splice(index, 1));
   };
 
-  useEffect(() => {
-    valueUpdated(dataArray.map((item) => item.val));
-  }, [dataArray]);
-
```
Ignusmart, line 37: This one was causing that default values were not parsed as expected on first render of the function input, getting wrong params values (into an array), this before the fix:

Screenshot 2024-04-06 at 11 07 27 AM

With this removal, here the right params values

Screenshot 2024-04-06 at 11 08 07 AM

packages/website/src/features/Packages/FunctionInput.tsx
```diff
@@ -33,8 +33,9 @@ export const FunctionInput: FC<{
   };
 
   useEffect(() => {
+    if (!isArray) return;
     valueUpdated(dataArray.map((item) => item.val));
-  }, [dataArray]);
+  }, [dataArray, isArray]);
```
Ignusmart, line 38: This one was causing that default values were not parsed as expected on first render of the function input, getting wrong params values (into an array), this before the fix:

Screenshot 2024-04-06 at 11 07 27 AM

With this update, here the right params values

Screenshot 2024-04-06 at 11 08 07 AM

packages/website/src/features/Packages/FunctionInput/TupleInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { FunctionInput } from '../FunctionInput';
+import { useState } from 'react';
+import { BigNumber } from 'ethers';
+import { AbiParameter } from 'viem';
+
+const TupleInput = ({
+  input,
+  handleUpdate,
+}: {
+  input: any;
```
mjlescano, line 11: Can this be typed as `input: AbiParameter;` ?

packages/website/src/features/Packages/FunctionInput/TupleInput.tsx
```diff
@@ -0,0 +1,70 @@
...
+  handleUpdate: (value: any) => void;
```
mjlescano, line 12: Also here `value: AbiParameter`

packages/website/src/features/Deploy/DisplayedTransaction.tsx
```diff
@@ -103,6 +104,47 @@ export function DisplayedTransaction(props: {
     return val.toString();
   }
 
+  function renderInput(type: string, val: string): React.ReactElement {
+    if (type === 'tuple') {
+      return (
+        <CopyBlock
```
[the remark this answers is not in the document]
noahlitvin: I think monaco is too "heavy" for this? Also don't like that we don't really have control over the scroll with it.

noahlitvin:
LGTM! Here's one edge case to consider: https://usecannon.com/packages/synthetix-omnibus/latest/8453-andromeda/interact/system/CoreProxy/0x32C222A9A159782aFD7529c87FA34b96CA72C696#selector-0x5d8c8844

packages/website/src/features/Packages/FunctionInput/TupleInput.tsx
Ignusmart: Good Catch! sure! Thanks @mjlescano

packages/website/src/features/Packages/FunctionInput/TupleInput.tsx
Ignusmart: Good Catch! sure! Thanks @mjlescano

noahlitvin (approved):

PR #883, opened 2024-04-10
fix(website): current safe chainId - queue cannon

Ignusmart:
The main purpose of this PR is to make sure to use the current safe chainId to look up for previous published package instead of using the current wallet provider chainId

https://github.com/usecannon/cannon/assets/3952453/a77d2370-f7bf-4c0f-9ded-3307ce25105b


mjlescano (approved):

dbeal-eth (approved):
good change. in the future we should probably automatically force the user's wallet to the correct network before they start trying to stage txns as well? tbh im not sure what the ideal ux is for this.

PR #886, opened 2024-04-11
fix(build): update build command to not write to remote ipfs

mjlescano:
This PR makes sure that the `Loader` used in the `publish` command does not write to remote ipfs node. On a future PR we can update all the other commands to not write to IPFS by default, except for publish, which is the only one that should do it.

Sorry for not adding a proper test case here, there are not tests on build command yet and it would take some time.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -130,9 +130,11 @@ export async function build({
     getSigner:
       getSigner ||
       async function (addr: viem.Address) {
+        const client = provider as unknown as viem.TestClient;
+
         // on test network any user can be conjured
-        await (provider as unknown as viem.TestClient).impersonateAccount({ address: addr });
-        await (provider as unknown as viem.TestClient).setBalance({ address: addr, value: viem.parseEther('10000') });
+        await client.impersonateAccount({ address: addr });
+        await client.setBalance({ address: addr, value: viem.parseEther('10000') });
 
         return {
           address: addr,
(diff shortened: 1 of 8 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -130,9 +130,11 @@ export async function build({
     getSigner:
       getSigner ||
       async function (addr: viem.Address) {
+        const client = provider as unknown as viem.TestClient;
+
         // on test network any user can be conjured
-        await (provider as unknown as viem.TestClient).impersonateAccount({ address: addr });
-        await (provider as unknown as viem.TestClient).setBalance({ address: addr, value: viem.parseEther('10000') });
+        await client.impersonateAccount({ address: addr });
+        await client.setBalance({ address: addr, value: viem.parseEther('10000') });
```
mjlescano, line 137: Just a type/visual fix

packages/cli/src/commands/build.ts
```diff
@@ -156,7 +158,12 @@ export async function build({
 
   const resolver = overrideResolver || (await createDefaultReadRegistry(cliSettings));
 
-  const runtime = new ChainBuilderRuntime(runtimeOptions, resolver, getMainLoader(cliSettings), 'ipfs');
+  const runtime = new ChainBuilderRuntime(
+    runtimeOptions,
+    resolver,
+    getMainLoader(cliSettings, { writeToIpfs: false }),
```
mjlescano, line 164: This is the main change to the `build` command.

FuzzB0t (approved):

FuzzB0t:
Looks like the some of the e2e tests are breaking due to a 'Missing IPFS loader' error. should be an easy fix but I'd say this looks good to go.

PR #890, opened 2024-04-12
feat(website): provisioned packages code

Ignusmart:
The purpose of this PR is to display code from provisioned packages in the Code tab.

https://github.com/usecannon/cannon/assets/3952453/dc71b78a-6679-4cc4-9dff-9960b865e3aa

### DIFF
--- a/packages/website/src/features/Packages/CodeExplorer.tsx
+++ b/packages/website/src/features/Packages/CodeExplorer.tsx
@@ -30,9 +30,64 @@ const handleDownload = (content: JSON, filename: string) => {
   URL.revokeObjectURL(url);
 };
 
+const PackageButton: FC<{
+  name: string;
+  selected: boolean;
+  onClick: () => void;
+}> = ({ name, selected, onClick }) => {
+  return (
+    <Button
+      color="white"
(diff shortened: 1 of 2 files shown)


FuzzB0t (approved):

packages/website/src/features/Packages/CodeExplorer.tsx
```diff
@@ -43,6 +98,41 @@ export const CodeExplorer: FC<{
...
+      : [];
+  const provisionArtifacts = provisionedPackagesKeys.map((k: string) => {
+    return {
+      name: k,
+      artifacts: (
+        deploymentData?.data?.state[`provision.${k}`] ||
+        deploymentData?.data?.state[`clone.${k}`]
+      )?.artifacts,
+    };
+  });
```
Ignusmart, line 118: @mjlescano could you confirm this is the intended behavior?

mjlescano: Yup, that looks right, so is still compatible with old packages and new ones, right?

mjlescano (approved):

PR #897, opened 2024-04-12
fix: on chain storage error

mjlescano:


packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -378,11 +373,7 @@ function QueueFromGitOps() {
     );
   }
 
-  if (
-    deployOnChainStore.isSuccess &&
-    !deployOnChainStore.isPending &&
-    !deployOnChainStore.isError
-  ) {
+  if (!deployOnChainStore.isSuccess && !deployOnChainStore.isPending) {
```
mjlescano, line 376: This is the fix, if there's an error when trying to interact with the contract on `deployOnChainStore`, show the message

noahlitvin (approved):

PR #899, opened 2024-04-15
fix(builder): var should detect changes to config

dbeal-eth:


noahlitvin:
Confirming it also detects changes to related vars set at in a clone step?

dbeal-eth:
> Confirming it also detects changes to related vars set at in a clone step? @dbeal-eth

if we wanted to be extra careful we should have a fully fledged e2e test verifying this kind of stuff, but for now the problem here was pretty fundamental so I would expect this to completely resolve the issue

PR #905, opened 2024-04-16
feat: code deep linking

Ignusmart:
This PR introduces deep linking functionality, allowing users to navigate directly from the `Interact` tab to the specific source code of a function. This feature is activated only when a corresponding source for the function is available.

https://github.com/usecannon/cannon/assets/3952453/8e57c02f-e370-43f3-aab8-9e43d4e29046


mjlescano (approved):
♥️

noahlitvin (approved):

PR #934, opened 2024-04-28
fix(registry): remove hardcoding of L1 chainId

mjlescano:
Changes necessary to fix registry deployment on Sepolia + OP Sepolia


dbeal-eth (approved):
Good work

Btw in the constructor, the optimism arg is misspelled. Maybe fix?

PR #953, opened 2024-05-01
feat(cli): Add unpublish command

saeta-eth:
This PR introduces the `unpublish` command to remove packages from the on-chain registry.

Usage:
- `cannon unpublish packageName`: Prompts the user to select a version and preset.
- `cannon unpublish packageName:version@preset`: Directly removes a specific deployment from the on-chain registry.


packages/builder/src/registry.ts
```diff
@@ -402,6 +443,18 @@ export class OnChainRegistry extends CannonRegistry {
...
+    console.log(bold(blueBright('\nUnpublishing package to the registry on-chain...\n')));
+    const packageData = this._preparePackageData(packagesNames, chainId, url, metaUrl);
+    return [await this._unpublishPackages([packageData])];
+  }
+
+  async unpublishMany(toUnpublish: { name: string[]; chainId: number; url: string; metaUrl?: string }[]): Promise<string[]> {
+    console.log(bold(blueBright('\nUnpublishing packages to the registry on-chain...\n')));
+    const packageDatas = toUnpublish.map((p) => this._preparePackageData(p.name, p.chainId, p.url, p.metaUrl));
+    return [await this._unpublishPackages(packageDatas)];
+  }
```
mjlescano, line 456: Looks like this fn was added to maintain consistency with `publish` & `publishMany`, but is not really necessary right now, because we always unpublish several tags but from the same packageName. So, this should be removed just to avoid extra code.

packages/builder/src/registry.ts
```diff
@@ -402,6 +443,18 @@ export class OnChainRegistry extends CannonRegistry {
...
+
```
mjlescano, line 457: This methods should not need to receive the `url` & `metaUrl` parameters, as they are not needed for calling the `unpublish` on the contract.

mjlescano (changes requested):

mjlescano (approved):

PR #960, opened 2024-05-02
fix(website): offer to switch

dbeal-eth:
works by always calling switch network prior to executing the actual transaction.

NOTE: if this seems to be working as expected, change the code in other places where we have buttons/etc that require switching network.

for exmaple:
- **when running a transaction on the explore tab** ✅ - Already asking user to switch to the right chain, there is an error showing up, but does not have to do with the switching chain action

https://github.com/usecannon/cannon/assets/3952453/d5e598df-f0c5-4c98-8644-dd6026efaeba

- **when staging a transaction on queue transactions** ✅ - This path is already covered, here a video showing what happens if user tries to queue a TX using a different chain for package / wallet or safe

https://github.com/usecannon/cannon/assets/3952453/be039ae5-144a-452c-b0a4-eb7291430e7c

- when signing a transaction on the gnosis safe 🟡 
- when staging a transaction on github 🟡

fixes https://linear.app/usecannon/issue/CAN-291/publish-to-optimism-button-doesnt-prompt-to-change-networks


dbeal-eth:
has anyone been able to verify if this is working? otherwise I will probably run into it this week when I am working on code on synthetix side.

PR #962, opened 2024-05-03
feat(metrics): basic prometheus metrics

mjlescano:
Implement basic Prometheus metrics on the API using https://www.npmjs.com/package/express-prom-bundle.


dbeal-eth (approved):
great work

PR #963, opened 2024-05-03
feat(api): improvements for search bar

mjlescano:
* Now, the api filters out all weird characters the client sends
* Improved packageName query and `-` compat. E.g. `reya-core` queries for `(@name:*reya* | @name:*core* | @name:%reya% | @name:%core%)`
* Results are sorted by string distance to the given query using [Levenshtein](https://www.npmjs.com/package/fastest-levenshtein)


dbeal-eth (approved):
amazing

PR #969, opened 2024-05-03
fix(builder): `clone` takes in explicit `target` option

dbeal-eth:
deprecates the `targetPreset` option. Now we have a lot of deprecated stuff in `clone` :laugh:

fixes https://linear.app/usecannon/issue/CAN-278/add-concise-target-prop-for-clone-operation-deprecating-the-others


FuzzB0t (approved):
lgtm

Issue #906, opened 2024-04-16
Basescan Verification Check


kaleb-keny:
basescan allows for only 100 contracts to be verified per day per api. However, synthetix v3 encompasses over 100 markets. Hence it is impossible to verify all the markets as cannon does not check whether a market is already verified before attempting verification. A request is to check whether verification is needed (which is api light) as opposed to attempting verify everything.
```
CANNON_ETHERSCAN_API_KEY=XXXXXXXXXXXXXXX cannon verify synthetix omnibus:latest@andromeda  --chain-id 84532
```

PR #996, opened 2024-05-08
feat(cli): check if contract is already verified

saeta-eth:
This PR wants to check if the contract is already verified before sending the request to verify it in Etherscan.

Closes https://github.com/usecannon/cannon/issues/906

### DIFF
--- a/packages/builder/src/package.ts
+++ b/packages/builder/src/package.ts
@@ -175,7 +175,10 @@ export async function getProvisionedPackages(packageRef: string, chainId: number
 
     return {
       packagesNames: _.uniq([def.getVersion(preCtx) || 'latest', ...(context && context.tags ? context.tags : tags)]).map(
-        (t: string) => `${def.getName(preCtx)}:${t}@${context && context.preset ? context.preset : preset || 'main'}`
+        (t: string) =>
+          // backwards compatibility: use target if its defined in context first; otherwise, revert to old way
+          (context && context.target) ||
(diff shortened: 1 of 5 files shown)


packages/cli/src/util/verify.ts
```diff
@@ -0,0 +1,70 @@
...
+    const json = (await response.json()) as EtherscanGetSourceCodeResponse;
+
+    if (json.message !== 'OK') {
+      return false;
+    }
+
+    const sourceCode = json.result[0]?.SourceCode;
+    return sourceCode !== undefined && sourceCode !== null && sourceCode !== '';
+  } catch (e) {
+    throw new Error('Failed to check if contract is verified on Etherscan: ' + (e as Error).message);
```
dbeal-eth, line 68: should explicitly return `false` after catching this error im pretty sure?

packages/cli/src/util/verify.ts
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Check if a smart contract is verified on Etherscan.
+ * @link https://docs.etherscan.io/api-endpoints/contracts#get-contract-source-code-for-verified-contract-source-codes
+ * @param address - The address of the smart contract.
+ * @param apiUrl - Etherscan API URL.
+ * @param apiKey - Etherscan API Key.
+ * @returns True if the contract is verified, false otherwise.
+ */
+
+export async function isVerified(address: string, apiUrl: string, apiKey: string) {
```
dbeal-eth, line 40: add explicit type to return boolean (see below comment)

dbeal-eth (approved):

PR #997, opened 2024-05-08
fix(builder): add backmerge of extras to settings

mjlescano:


dbeal-eth (changes requested):
please update the `builder.test.ts` unit tests to verify correct handling of extras and vars going both directions, and ideally if possible, nested (could  be as simple as adding checks/conditions). alternatively, update e2e tests to explicitly verify this condition.

PR #1011, opened 2024-05-12
feat(website): payable functions - payment input

Ignusmart:
The main purpose of this PR is to add a payment input for payable functions.

https://github.com/usecannon/cannon/assets/3952453/e8e9c49f-e54e-46cf-b362-0f8f707b4018

### DIFF
--- a/packages/website/src/features/Deploy/QueueDrawer.tsx
+++ b/packages/website/src/features/Deploy/QueueDrawer.tsx
@@ -330,7 +330,7 @@ export const QueuedTxns = ({
                     })
                   }
                 />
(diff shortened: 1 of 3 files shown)


dbeal-eth (approved):
looks good. though, I think it would be more natural in most cases to accept user input the opposite way (ie, user types in the payment amount in ETH, and then the display shows that value in wei)

noahlitvin:
Let's replace `Payment (WEI)` with `Value` and put `wei` in a `RightAddon` https://v2.chakra-ui.com/docs/components/input/usage#left-and-right-addons

PR #1012, opened 2024-05-12
fix(website): function array outputs render

Ignusmart:
Previously, function call outputs were duplicated, rendering the same values in each output field.

image

- With the fix, the function call outputs are now being rendered as expected.

https://github.com/usecannon/cannon/assets/3952453/7d749a2e-45ac-46ad-8658-9bac96208231

- Functions which return tuples, or arrays of tuples (such as https://usecannon.com/packages/synthetix/latest/10-main/interact/synthetix/CoreProxy/0xffffffaEff0B96Ea8e4f94b2253f31abdD875847#selector-0x75bf2444 ) working as expected

https://github.com/usecannon/cannon/assets/3952453/ac48cc4a-4aea-42a1-b1c3-d450582b7321

### DIFF
--- a/packages/website/src/features/Packages/FunctionOutput.tsx
+++ b/packages/website/src/features/Packages/FunctionOutput.tsx
@@ -36,7 +36,11 @@ export const FunctionOutput: FC<{
     </Box>
   );
 
-  const renderOutput = (item: AbiParameter, value: { [key: string]: any }) => {
+  const renderOutput = (
+    item: AbiParameter,
+    value: { [key: string]: any },
+    index?: number
+  ) => {
     if (item.type === 'tuple' && hasComponents(item) && value) {
       return (
         <Box pl="4">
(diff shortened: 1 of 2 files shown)


packages/website/src/features/Packages/FunctionOutput.tsx
```diff
@@ -82,11 +86,19 @@ export const FunctionOutput: FC<{
...
+        if (index !== undefined) {
+          return (
+            <Text fontSize="xs" display="block">
+              {String(value[index])}
+            </Text>
+          );
+        } else {
+          return value.map((val, idx) => (
+            <Text fontSize="xs" display="block" key={idx}>
+              {index !== undefined ? String(val[index]) : String(val)}
```
mjlescano, line 98: In this case, `index !== undefined` is giving always `false`, right?

dbeal-eth:
@Ignusmart looks good. im guessing you checked other functions which return tuples, or arrays of tuples (such as https://usecannon.com/packages/synthetix/latest/10-main/interact/synthetix/CoreProxy/0xffffffaEff0B96Ea8e4f94b2253f31abdD875847#selector-0x75bf2444 )



we have had issues with updating these functions in the past where we fix one thing and break another.

packages/website/src/features/Packages/FunctionOutput.tsx
Ignusmart: Hey @mjlescano ! you're right! thanks for catching this one

noahlitvin (approved):

PR #1019, opened 2024-05-14
feat(cli): skip pulling sources from foundry if not public source code

dbeal-eth:
not including sources would significantly improve foundry build times by skipping the IPC calls to foundry entirely.


dbeal-eth:
client confirmed that this change works as expected for their needs

mjlescano (approved):

PR #1021, opened 2024-05-14
fix(website): offer to switch to right safe chain

Ignusmart:
The main purpose of this PR is to offer the user to switch to the right chain before executing/staging any transaction

https://github.com/usecannon/cannon/assets/3952453/eaf6339a-a36a-4ca7-b512-492498e2b1ba

### DIFF
--- a/packages/website/src/features/Deploy/QueueDrawer.tsx
+++ b/packages/website/src/features/Deploy/QueueDrawer.tsx
@@ -43,7 +43,7 @@ import {
   TransactionRequestBase,
   zeroAddress,
 } from 'viem';
-import { useAccount, useWriteContract, useSwitchChain } from 'wagmi';
+import { useAccount, useWriteContract } from 'wagmi';
 import NoncePicker from './NoncePicker';
 import { QueueTransaction } from './QueueTransaction';
 import { SafeAddressInput } from './SafeAddressInput';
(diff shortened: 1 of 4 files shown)


dbeal-eth (changes requested):
in the video you shared, the button appears to un-disable after the switch chain is completed? Shouldnt it continue to be disabled/loading until its done?

also, can we turn this more or less repeated section of code into a helper function? I have a feeling we are going to be using similar pattern a lot.

packages/website/src/features/Deploy/QueueDrawer.tsx
```diff
@@ -380,12 +380,15 @@ export const QueuedTxns = ({
                           colorScheme="teal"
                           w="100%"
                           isDisabled={
+                            stager.signing ||
```
dbeal-eth, line 383: in accordance with showing a reason any time a button is disabled, can we put some indication somewhere... somehow on the page that says "currently signing txn" to indicate the button is disabled (maybe the user doenst realize their signing window was opened when they click on the button)

dbeal-eth (approved):
i still have an umber of things that I think could be improved about this, but we have to set this PRs stopping point somewhere.

packages/website/src/features/Deploy/QueueDrawer.tsx
Ignusmart: Nice catch, sure, I'll add it ASAP, thank you

dbeal-eth (approved):

PR #1022, opened 2024-05-15
fix(website): default nonce to 1

mjlescano:


mjlescano:
Closed in favor or https://github.com/usecannon/cannon/pull/1020/files

PR #1024, opened 2024-05-15
feat(website): set up pages router to work with 100% client side rendering

dbeal-eth:


Ignusmart (approved):
Overall, the code looks good. We'll need to use the app to check for any regressions, but I believe everything is fine. ILGTM 🚀

PR #1028, opened 2024-05-16
feat(cli): register packages in batches

saeta-eth:
This PR aims to add the ability to register more than one package at a time.

Usage: `cannon register my-package my-another-package`

### DIFF
--- a/packages/cli/src/commandsConfig.ts
+++ b/packages/cli/src/commandsConfig.ts
@@ -8,15 +8,15 @@ const debugVerbosity = [
   {
     flags: '-vv',
     description:
-      'Print logs for builder and its definition section, equivalent to DEBUG=cannon:builder,cannon:builder:definition',
+      'Print logs for builder and its definition section, equivalent to DEBUG=cannon:builder, cannon:builder:definition',
   },
   {
     flags: '-vvv',
-    description: 'Print logs for builder and its all sub sections,equivalent to DEBUG=cannon:builder*',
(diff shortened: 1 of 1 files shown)


packages/builder/src/registry.ts
```diff
@@ -700,13 +664,46 @@ export class OnChainRegistry extends CannonRegistry {
...
+          setAdditionalPublishersTxs.push(txn);
+          break;
+        }
+        default:
+          throw new Error(`Unknown function name: ${txn.functionName}`);
+      }
+    }
+
+    return prepareMulticall([...nominatePackageOwnerTxs, ...setPackageOwnershipTxs, ...setAdditionalPublishersTxs]);
+  }
```
mjlescano, line 698: Unless it is absolutely necessary, this function for reordering the calls is not necessary.

packages/cli/src/commands/register.ts
```diff
@@ -39,22 +41,29 @@ export async function register({ cliSettings, options, packageRef, fromPublish }
...
+        pkg.name,
+        optimismRegistryConfig.address
+      );
+
+      if (isRegisteredOnMainnet && isRegisteredOnOptimism) {
+        throw new Error(`The package "${pkg.name}" is already registered.`);
+      }
+
+      return [isRegisteredOnMainnet, isRegisteredOnOptimism];
+    })
```
mjlescano, line 64: The `isPackageRegistered` function could be improved to be able to take one `contractAddress` per provider, and this block be removed almost completely. (it would also parallelize both calls and have a small speed improvement)

mjlescano (approved):

PR #1029, opened 2024-05-16
fix(website): add packageName when enqueuing transactions

mjlescano:
This PR proposed a new title for each enqueued TX, to show the package name.

This is specially useful when coming back from a previously saved state, if I don't remember from which package was created the TX, now it can be checked. 

(also updates the X icon for a 🗑️ )

Before:
Screenshot 2024-05-16 at 20 36 06

After:
Screenshot 2024-05-16 at 20 36 21

### DIFF
--- a/packages/website/src/features/Deploy/QueueDrawer.tsx
+++ b/packages/website/src/features/Deploy/QueueDrawer.tsx
@@ -222,7 +222,6 @@ export const QueuedTxns = ({
         {queuedIdentifiableTxns.length > 0
           ? queuedIdentifiableTxns.map((queuedIdentifiableTxn, i) => (
               <Box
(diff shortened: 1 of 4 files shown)


packages/website/src/features/Deploy/QueueDrawer.tsx
```diff
@@ -222,9 +222,10 @@ export const QueuedTxns = ({
         {queuedIdentifiableTxns.length > 0
           ? queuedIdentifiableTxns.map((queuedIdentifiableTxn, i) => (
               <Box
+                overflow="hidden"
```
mjlescano, line 225:
```suggestion
```

Ignusmart (approved):
🚀

PR #1036, opened 2024-05-18
fix(website): decode staged txns / support multipackage

Ignusmart:
The main purpose of this PR is to ensure that queued transactions are decoded as expected when they are queued using an existing package (from the interact page, or from the drawer / stage page). Additionally, this PR adds support for staging multi-package transactions.

## Queuing txs from the same package (from the drawer and interact page)

https://github.com/usecannon/cannon/assets/3952453/2b1c569d-a5ee-41c6-afb9-480bd4c1a02d

## Queueing txs from different packages

https://github.com/usecannon/cannon/assets/3952453/e90385eb-a99f-495b-9245-d1944a08a3d2


dbeal-eth (approved):
looks like excellent work

PR #1047, opened 2024-05-21
fix(website): performance improvments when rendering txs

mjlescano:
https://github.com/usecannon/cannon/assets/3952453/5e2b8359-bb3b-4c3b-91be-932307681510

### DIFF
--- a/packages/website/src/features/Deploy/TransactionDisplay.tsx
+++ b/packages/website/src/features/Deploy/TransactionDisplay.tsx
@@ -15,6 +15,7 @@ import {
   AlertTitle,
   Box,
   Flex,
+  Image,
   Link,
   Modal,
   ModalBody,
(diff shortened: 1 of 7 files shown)


packages/website/src/features/Deploy/TransactionDisplay.tsx
```diff
@@ -285,6 +285,8 @@ export function TransactionDisplay(props: {
                 pkgUrl={
                   hintData.isSinglePackage ? hintData.cannonPackage : pkgUrl
                 }
+                cannonInfo={cannonInfo}
+                isPreloaded={hintData.isSinglePackage}
```
mjlescano, line 289: In the case that is not a single package, it would still have performance issues?

Ignusmart: yes 😞 , I understand this is not too common though (to queue a lot of txs from different packages), but we can let it that way and think in another solution for the multipackage, something we can do is to fetch the data for unique packages and then pass the information to the corresponding txs, I mean, to fetch just the necessary data and not fetch data for each txn

Ignusmart (approved):

PR #1049, opened 2024-05-21
fix(api): gentler parsing of redis results

mjlescano:
This PR makes sure that the API does not return a `500` in case it cannot parse a result from the database. It just ignores the result and `console.warn` it so it could be potentially parsed in the future.

As an extra, adds the possibility to set a custom API endpoint, mostly for dev/test purposes.

### DIFF
--- a/packages/api/src/config.ts
+++ b/packages/api/src/config.ts
@@ -1,4 +1,5 @@
 import { cleanEnv, EnvError, makeValidator, str } from 'envalid';
+import 'dotenv/config';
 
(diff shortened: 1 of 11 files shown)


packages/api/src/config.ts
```diff
@@ -1,4 +1,5 @@
 import { cleanEnv, EnvError, makeValidator, str } from 'envalid';
+import 'dotenv/config';
```
mjlescano, line 2: Added it for easier local setup

PR #1050, opened 2024-05-21
fix(indexer): trim contract address

saeta-eth:


FuzzB0t (approved):

dbeal-eth (changes requested):
to be effective, also need to set the address stored in the actual hash (a few lines lower than you already edited)

mjlescano:
@dbeal-eth could you re-review this one? 🙏

dbeal-eth (approved):

PR #1054, opened 2024-05-22
fix(website): Add tooltips to preview Tx button

FuzzB0t:


packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -369,6 +369,64 @@ function QueueFromGitOps() {
...
+      </Tooltip>
+    );
+  }
+
+  function RenderPreviewButtonTooltip() {
+    if (settings.isIpfsGateway) {
+      return (
+        <PreviewButton message="Cannot write to IPFS endpoint as gateway, please change the IPFS Url in your settings." />
+      );
+    }
```
mjlescano, line 401: I think all the logic for `isIpfsGateway` can be removed, as it always returns `false` now: https://github.com/usecannon/cannon/blame/main/packages/builder/src/ipfs.ts/#L44-L47

mjlescano (approved):

packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
FuzzB0t: Oh right! Ok I'll remove this :)

PR #1057, opened 2024-05-26
fix(website): wallet disconnected after reload / navigate

Ignusmart:
This PR aims to solve the issue when wallet gets disconnected after navigate / reload the page, this one is a regression of this other PR https://github.com/usecannon/cannon/pull/1009

## Issue Video

Wallet gets disconnected after reload

https://github.com/usecannon/cannon/assets/3952453/965d2758-393a-4979-bcdb-242875699eb6

## With the fix

Wallet does not get disconnected after reload

https://github.com/usecannon/cannon/assets/3952453/0330befc-b371-4cf9-932b-f334f442d2cc

## With the fix in Opera Browser

https://github.com/usecannon/cannon/assets/3952453/a0c9145d-e41c-477d-aa6b-65ef9f82c0ae


FuzzB0t (approved):

Ignusmart:
After digging deeper it seems the transports are not the root cause of the issue, for that reason we are closing this one in order to open another PR trying to discover what the root cause is

PR #1060, opened 2024-05-27
fix(website): seo metadata

Ignusmart:
This PR aims to fix website SEO metadata after changing from APP dir feature to Pages structure


packages/website/src/pages/packages/[name]/index.tsx
```diff
@@ -11,32 +12,21 @@ const NoSSR = dynamic(
...
-  }*/
-
 export default function Deployment() {
   const params = useRouter().query;
-  return <NoSSR name={params.name as string} />;
+  return (
+    <>
+      <NextSeo
+        {...defaultSEO}
+        title={`Cannon | ${params.name}`}
```
Comment (dbeal-eth) [inline packages/website/src/pages/packages/[name]/index.tsx:21]: can we do this same pattern for all of the pages under the `packages`? Right now a lot of the pages are just "Package" when we should probably be embedding the params name like we are here

dbeal-eth (changes requested):

dbeal-eth (approved):

PR #1067, opened 2024-05-28
fix(website): Prevent FOUC

noahlitvin:
Unsure if we can do better than this without SSR?


dbeal-eth (approved):
we can do better without SSR because (for example) the front page never changes whati t displays 

so in theory, you should just be albe to send the static html of the front page post-render and it will look correct.

but my guess is we have a bunch of `NoSSR` sections which are affecting this process. We can optimize later

PR #1070, opened 2024-05-29
fix(cli): pin packages to ipfs

saeta-eth:
This PR aims to fix the issue that some users mentioned about not being able to pin to IPFS


dbeal-eth (approved):
paired on this, looks great

PR #1084, opened 2024-05-31
fix(website): show error when payable value is not valid

Ignusmart:
This PR aims to prevent client-side crashes due to invalid big integers entered into the payable function's value. Instead, a FormHelper text with an error message is rendered.

[Screencast from 2024-05-30 21-36-28.webm](https://github.com/usecannon/cannon/assets/3952453/dd8b90c3-2761-4f75-8726-e44e1dd61ef0)


FuzzB0t (approved):

noahlitvin (changes requested):
Maybe just `isInvalid` in on the input (red border) instead of `FormHelperText`?

Can you elaborate how we'd replace the Form Helper Text with `isInvalid`?

FuzzB0t:
> Maybe just `isInvalid` in on the input (red border) instead of `FormHelperText`?

Can you elaborate how we'd replace the Form Helper Text with `isInvalid`?

PR #1097, opened 2024-06-04
feat(cli): CAN-340 Show step name on error

noahlitvin:
Needs more testing/verification

### DIFF
--- a/packages/builder/src/builder.ts
+++ b/packages/builder/src/builder.ts
@@ -88,6 +88,9 @@ ${printChainDefinitionProblems(problems)}`);
     } else {
       debug('building individual');
       doActions: for (const n of topologicalActions) {
+        const [type, label] = n.split('.') as [keyof typeof ActionKinds, string];
+        runtime.emit(Events.PreStepExecute, type, label, {}, 0);
+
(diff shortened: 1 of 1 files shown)


noahlitvin:
Yea, I think the improvement is surfaced if e.g. it throws an error parsing a template string? (Like see what follows the log output when it's in the new position from this pr)

mjlescano:
This PR shoed no difference on these three cases (first two we can still see the name on `main`):
1. When doing something invalid on templating:
Screenshot 2024-07-05 at 14 44 42

2. When throwing an error on constructor:
Screenshot 2024-07-05 at 14 47 48

3. When adding an invalid depends, this case didn't show the step name on any branch:
Screenshot 2024-07-05 at 14 50 16

Case `3.` is the one we should be fixing probably, maybe am i missing some other case?

dbeal-eth (changes requested):
the problem with this implementation is that it prints out the step name every single time, even if its not actually being executed

in order to determine whether a step should be executed or not, its configuration needs to be computed, which requires doing some things which may result in error.

I think the only approach we have to this is adding a fake log message for this if an error is generated and only in that case.

PR #1098, opened 2024-06-06
feat(cli): read and write ipfs url for decentralization

dbeal-eth:
note: publishipfsurl/cannon repo is still separate. and optional if not publishing.

### DIFF
--- a/packages/cli/src/settings.ts
+++ b/packages/cli/src/settings.ts
@@ -37,9 +37,10 @@ export type CliSettings = {
   ipfsTimeout?: number;
 
   /**
+
    * the url of the IPFS endpoint to use as a storage base, only for reading data
    */
-  readIpfsUrl?: string;
+  ipfsUrl?: string;
 
   /**
    * the url of the IPFS endpoint to use as a storage base, only for writing data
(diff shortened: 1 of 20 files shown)


packages/cli/src/settings.ts
```diff
@@ -37,9 +37,14 @@ export type CliSettings = {
   ipfsTimeout?: number;
 
   /**
-   * the url of the IPFS endpoint to use as a storage base. defaults to localhost IPFS
+   * the url of the IPFS endpoint to use as a storage base, only for reading data
    */
-  ipfsUrl?: string;
```
noahlitvin, line 42: I thought the idea is we leave this and deprecate it? I think it'd be a fallback for readIpfsUrl?

noahlitvin:
Question on fallback behavior:

I think for read, we'd want to try filesystem/cache, `readIpfsUrl` (if set), also `ipfsUrl` (deprecated, but if present), and also default repo url?

For write, it goes to local cache, and writes to `writeIpfsUrl` if present (*not* the deprecated `ipfsUrl`)

For publish, it goes to `publishIpfsUrl` if present, otherwise the default repo url. Should it also write to the repo url when the `publishIpfsUrl` is set to something?

packages/cli/src/settings.ts
dbeal-eth: really like to avoid keeping it around if we are taking this direction

dbeal-eth:
> ant to try files

everything that you say here should be what its doign with this change, minus the existance of `ipfsUrl` setting

FuzzB0t (approved):

PR #1099, opened 2024-06-06
feat(cli): updates

noahlitvin:


packages/cli/src/settings.ts
```diff
@@ -36,6 +36,11 @@ export type CliSettings = {
    */
   ipfsTimeout?: number;
 
+  /**
+   * DEPRECATED: fallback for readIpfsUrl
```
mjlescano, line 40:
```suggestion
   * @deprecated: fallback for readIpfsUrl
```

dbeal-eth (approved):
Technically this doesn't handle the case where user has a config file with `ipfsUrl` defined it looks like, only the envvar

noahlitvin:
Can we just leave readIpfsUrl as ipfsUrl?

packages/cli/src/loader.test.ts
```diff
@@ -53,7 +53,7 @@ describe('getMainLoader', function getMainLoaderTestCases() {
     const settings: CliSettings = {
       providerUrl: '',
       registries: [],
-      readIpfsUrl: 'ipfs',
+      ipfsUrl: 'ipfs',
```
noahlitvin, line 56: This is what's on main? Seems weird it's just `ipfs` here?

PR #1105, opened 2024-06-10
chore: add Interact e2e tests for website

FuzzB0t:
Still missing write function tests, as there are some blockers when trying to test out wallet functionality + successfull write execution


noahlitvin:
Do we want these checking "latest" or a pinned version?

PR #1106, opened 2024-06-10
fix(website): misc fixes website

dbeal-eth:
* publishes different packages on different chains due to registry priority messup
* goes to correct explorer for the chain if user clicks wallet link
* apply height to header to prevent cls
* improve front page load speed by delayed loading for scroll down components (which happen to be very heavy due to special libs)


packages/website/src/features/Deploy/PublishUtility.tsx
```diff
@@ -78,27 +79,35 @@ export default function PublishUtility(props: {
...
-  const etherscanUrl =
-    findChain(props.targetChainId).blockExplorers?.default?.url ??
-    'https://etherscan.io';
-
   const { transports } = useProviders();
 
-  const prepareAndPublishPackage = async (registryChainId: number) => {
+  const prepareAndPublishPackage = async (
+    registryChainId: number,
+    roChainId: number
```
noahlitvin, line 86: What does `roChainId` mean here?

noahlitvin (approved):

packages/website/src/features/Deploy/PublishUtility.tsx
dbeal-eth: "read only chain id"

added it because even though publishing on one chain, it is stil necessary to read from the other chain to determine if package is published or not

not a perfect fix but if people are predominantly publishing to one chain its a good solution

PR #1113, opened 2024-06-10
fix(website): header minHeight

noahlitvin:


dbeal-eth (changes requested):
this doesnt effectively fix the issue because the problem is that the `marginBottom` is required to ensure the conents of the page do not shift due to breakpoint changes.

dbeal-eth (approved):
excellent!

PR #1115, opened 2024-06-12
feat(cli): add e2e tests for `register` and `publishers` commands

saeta-eth:
This PR introduces e2e tests for the register command. The tests cover the entire process, starting from the transaction executed on Anvil Mainnet to the successful propagation on the Anvil Optimism.

------

Update: E2E tests for the publishers command have been merged into this PR. https://github.com/usecannon/cannon/pull/1124


packages/cli/src/commands/publishers.ts
```diff
@@ -157,16 +183,16 @@ export async function publishers({ cliSettings, options, packageRef }: Params) {
...
+      type: 'confirm',
+      name: 'confirmation',
+      message: 'Proceed?',
+    });
 
-  if (!verification.confirmation) {
-    console.log('Cancelled');
-    process.exit(1);
+    if (!confirm.confirmation) {
+      process.exit(0);
```
mjlescano, line 194: The error code should be different to `0`, to mark the script as not completed or intervened by the user

noahlitvin (approved):

PR #1116, opened 2024-06-13
feat(builder): add strict validation to steps attributes

mjlescano:
Closes https://linear.app/usecannon/issue/CAN-348/throw-an-error-if-you-have-stuff-in-your-operation-definition-thats

This PR does not allow to add invalid config values to step names, for example in `[deploy.Contract]` it would throw if you misspell `artifact =` as `artifacttt =`.

### DIFF
--- a/examples/sample-hardhat-project/package-lock.json
+++ b/examples/sample-hardhat-project/package-lock.json
@@ -28,13 +28,13 @@
       }
     },
     "../../packages/hardhat-cannon": {
-      "version": "2.15.1",
+      "version": "2.15.2",
       "dev": true,
       "license": "MIT",
       "dependencies": {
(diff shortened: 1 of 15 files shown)


FuzzB0t (approved):

dbeal-eth (approved):

mjlescano:
The e2e tests are failing because the [`Synthetixio/synthetix-sandbox/cannonfile.toml#L119-L120`](https://github.com/Synthetixio/synthetix-sandbox/blob/main/cannonfile.toml#L119-L120) file has an error, using `extras.* = *` instead of `extra.* = *`. This is a good thing! But just to note that after we merge this PR all the existing cannon packages with typos are going to stop working.

PR #1128, opened 2024-06-17
fix(builder): misc

dbeal-eth:
quite a few misc fixes in here:
* add timestamp and block number info to more places
* fix(builder): all artifacts for trace 
* fix access recorder not properly resolving with steps that have logic
* better stack tracing
* better printing out of step name prior to printing out errors


mjlescano (approved):

noahlitvin:
Just rebased to this branch here: https://github.com/usecannon/cannon/pull/1097

PR #1133, opened 2024-06-18
fix(website): sync safe selector with url params

saeta-eth:
When users navigate to the deploy page, the safe is not automatically selected based on the URL parameters, resulting in an empty page. This PR addresses the issue.

Bug:
Screenshot 2024-06-18 at 11 41 14 AM


packages/website/src/features/Deploy/SafeAddressInput.tsx
```diff
@@ -56,42 +58,43 @@ export function SafeAddressInput() {
 
   // Load the safe address from url
   useEffect(() => {
-    if (searchParams.address || searchParams.chainId) {
-      const chainId = searchParams.chainId;
-      const address = searchParams.address;
-      const newSafe = parseSafe(`${chainId}:${address}`);
+    const { safeAddress: address, chainId } = searchParams;
+
+    if (isNaN(Number(chainId))) return;
```
mjlescano, line 63: On this case is the same, but I usually prefer the `isSafeInteger` check, as it also checks for undefined, weird floats, etc
```suggestion
    if (!Number.isSafeInteger(Number(chainId))) return;
```

noahlitvin (approved):

PR #1145, opened 2024-06-21
fix(website): code linking

saeta-eth:
This PR aims to fix an issue where the code linking feature was not working properly because the URL was generated incorrectly.

Closes https://linear.app/usecannon/issue/CAN-338/function-code-button-link-is-not-working


noahlitvin:
Seems like its stripping off the function name? e.g. at `/packages/synthetix-omnibus/latest/8453-andromeda/code/system?source=contracts%2Fmodules%2Faccount%2FAccountTokenModule.sol&function=mint`

noahlitvin (approved):

PR #1175, opened 2024-06-29
fix(cli): Censor ipfs password

FuzzB0t:
Adds a filter for the password to the ipfs urls on the build command
image


packages/cli/src/commands/build.ts
```diff
@@ -409,6 +409,9 @@ export async function build({
     process.off('SIGTERM', handler);
     process.off('SIGQUIT', handler);
 
+    const filteredWriteIpfsUrl = cliSettings.writeIpfsUrl?.replace(RegExp(/:[a-zA-Z0-9]+@/), `:${'*'.repeat(20)}@`);
+    const filteredPublishIpfsUrl = cliSettings.publishIpfsUrl?.replace(RegExp(/:[a-zA-Z0-9]+@/), `:${'*'.repeat(20)}@`);
```
mjlescano, line 413:
```suggestion
    const filteredWriteIpfsUrl = cliSettings.writeIpfsUrl?.replace(/:[a-zA-Z0-9]+@/, `:${'*'.repeat(20)}@`);
    const filteredPublishIpfsUrl = cliSettings.publishIpfsUrl?.replace(/:[a-zA-Z0-9]+@/, `:${'*'.repeat(20)}@`);
```

packages/cli/src/commands/build.ts
```diff
@@ -409,6 +409,9 @@ export async function build({
     process.off('SIGTERM', handler);
     process.off('SIGQUIT', handler);
 
+    const filteredWriteIpfsUrl = cliSettings.writeIpfsUrl?.replace(RegExp(/:[a-zA-Z0-9]+@/), `:${'*'.repeat(20)}@`);
```
mjlescano, line 412: This Regex does not work when the password has special characters like `!-_`, nevertheless, it would be easier to use the native Node's URL class, e.g.:
```typescript
const filterUrlPassword = (uri: sting) => {
  const res = new URL(uri);
  res.password = '*'.repeat(10);
  return res.string;
}

// ...

const filteredWriteIpfsUrl = filterUrlPassword(cliSettings.writeIpfsUrl)
```

FuzzB0t: Thanks for the review! I've refactored the code to use the internal URL lib.

mjlescano (approved):

PR #1185, opened 2024-07-02
fix(builder): Add packageref validation for name and variant size

FuzzB0t:
Added a validation on the package reference class that throws an error when name or variant size exceeds 32 bytes
image

fixes https://linear.app/usecannon/issue/CAN-376/verify-target-in-clone-step-isnt-too-long-to-publish-before-building


packages/builder/src/package.ts
```diff
@@ -89,6 +89,17 @@ export class PackageReference {
   }
 
   static isValid(ref: string) {
+    const pkgRef = new PackageReference(ref);
```
FuzzB0t, line 92: I added the error logic here so that I could run this function on the zod schema as a refine, however it feels wrong to add error logic to a boolean function. There's likely a better way to achieve this.

FuzzB0t: Refactored this to add the error logic in the parse function of the package ref class, and add custom refinements in zod on the schema itself.

dbeal-eth (dismissed):
Looking good. We definitely want to add some tests for this. Also, some of the checks appear to be validating from `name:version` pair, only the name, but the version is subject to 32 length as well.

FuzzB0t:
> Looking good. We definitely want to add some tests for this. Also, some of the checks appear to be validating from `name:version` pair, only the name, but the version is subject to 32 length as well.

Agree on the tests, will add them asap.
Which checks are you referring to? 
The checks I added check to make sure the name isn't more than 32 bytes and that the version+preset (aka variant) isn't more than 32 bytes as that's what gets passed into the registry on publish.

packages/builder/src/steps/pull.ts
```diff
@@ -9,12 +9,12 @@ import { ChainBuilderRuntime } from '../runtime';
...
 
-const debug = Debug('cannon:builder:import');
+const debug = Debug('cannon:builder:pull');
 
 /**
- *  Available properties for import operation
+ *  Available properties for pull operation
  *  @public
- *  @group Import
+ *  @group pull
```
mjlescano, line 17: ❤️

packages/builder/src/package.ts
```diff
@@ -37,7 +37,7 @@ export interface PackagePublishCall {
 export class PackageReference {
   static DEFAULT_TAG = 'latest';
   static DEFAULT_PRESET = 'main';
-  static PACKAGE_REGEX = /^(?<name>@?[a-z0-9][A-Za-z0-9-]{1,30}[a-z0-9])(?::(?<version>[^@]+))?(@(?<preset>[^\s]+))?$/;
+  static PACKAGE_REGEX = /^(?<name>@?[a-z0-9][A-Za-z0-9-]{1,}[a-z0-9])(?::(?<version>[^@]+))?(@(?<preset>[^\s]+))?$/;
```
mjlescano, line 40: The `PackageReference.isValid` fn now only uses the regex to validate, not the constructor, so it would not return `false` when given a too long package name.

packages/builder/src/package.ts
```diff
@@ -76,12 +76,23 @@ export class PackageReference {
...
       throw new Error(
-        `Invalid package name "${ref}". Should be of the format <package-name>:<version> or <package-name>:<version>@<preset>`
+        `Invalid package reference "${ref}". Should be of the format <package-name>:<version> or <package-name>:<version>@<preset>`
       );
     }
 
     const res: PartialRefValues = { name: match.groups.name };
 
+    const nameSize = new Blob([res.name]).size;
+    const variantSize = new Blob([res.version + '_' + res.preset]).size;
```
mjlescano, line 86: no need to use `Blob.size` here, as the regex makes sure that there are no values that occupy more than 1 char because the only possible chars are `[a-z0-9-_]`, so `res.name.length` should work correctly AFAIK

packages/builder/src/schemas.ts
```diff
@@ -536,7 +591,34 @@ export const cloneSchema = z
...
+
+              if (match) {
+                const versionSize = new Blob([match!.groups!.version]).size;
+
+                return versionSize <= 32;
+              } else {
+                return true;
+              }
+            },
+            (val) => ({ message: `Package reference "${val}" is too long. Package version exceeds 32 bytes` })
```
mjlescano, line 619: These functions should not be repeated, maybe they can be refactored as a single new fn in `PackageReference` that's called `validate` and returns `undefined` in case the given ref is correct, or an error string in case it found one.

mjlescano: Then, that `PackageReference.validate` fn can be used by `PackageReference.isValid`, `PackageReference.constructor` and also the schema validators

packages/builder/src/package.ts
FuzzB0t: Yeah im only using the isValid function to check whether it matches the format, however when setting the length inside the regex it does fail on `.match` due to length

packages/builder/src/schemas.ts
FuzzB0t: There's an issue when importing from package through the schema file, which causes the action to register incorrectly, which is why I omitted using it in the schema section alltogether.

FuzzB0t: However I could refactor them into a single function that takes a value and checks the length to be less or equal to 32 bytes, so that we avoid repetition.

packages/builder/src/package.ts
mjlescano: Agree on the `parse` functionality with prettier errors, but the `PackageReference.isValid` should return `false` when it is too long, and with this change it doesn't. Maybe instead of validating with the regex inside the `isValid` you can use `.parse` with a try catch.

packages/builder/src/package.ts
```diff
@@ -76,19 +76,35 @@ export class PackageReference {
...
     return res;
   }
 
   static isValid(ref: string) {
+    try {
+      PackageReference.parse(ref);
+    } catch (err) {
+      return false;
+    }
     return !!PackageReference.PACKAGE_REGEX.test(ref);
```
mjlescano, line 108: The regex is already being used inside the `.parse`:
```suggestion
    try {
      PackageReference.parse(ref);
      return true;
    } catch (err) {
      return false;
    }
```

mjlescano (approved):

PR #1186, opened 2024-07-05
fix(builder): ability to handle nested helper context

dbeal-eth:
access recorder can incorrectly handle some CannonHelperContext helpers if they are nested in an object

to fix this, update the function to properly handle nested objects.

tested on actual issue occuring on synthetix v3 repository

### DIFF
--- a/packages/builder/src/access-recorder.ts
+++ b/packages/builder/src/access-recorder.ts
@@ -70,6 +70,15 @@ export function computeTemplateAccesses(str?: string, possibleNames: string[] = 
     if (typeof (CannonHelperContext as any)[n] === 'function') {
       // the types have been a massive unsolvableseeming pain here
       recorders[n] = _.noop as unknown as AccessRecorder;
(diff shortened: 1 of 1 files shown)


packages/builder/src/access-recorder.ts
```diff
@@ -70,6 +70,15 @@ export function computeTemplateAccesses(str?: string, possibleNames: string[] =
     if (typeof (CannonHelperContext as any)[n] === 'function') {
       // the types have been a massive unsolvableseeming pain here
       recorders[n] = _.noop as unknown as AccessRecorder;
+    } else if (typeof (CannonHelperContext as any)[n] === 'object') {
+      for (const o in (CannonHelperContext as any)[n]) {
+        (recorders[n] as any) = {};
+        if (typeof (CannonHelperContext as any)[n][o] === 'function') {
```
mjlescano, line 76: For these comparisons recommend to use the `_.isFunction`, `_.isPlainObject` and `_.isObject` fns

mjlescano (approved):

PR #1192, opened 2024-07-09
feat(cli): add hideApiKey helper function

saeta-eth:
Bug: We're showing the api key of the provider:

Screenshot 2024-07-08 at 9 26 03 PM

Also, the current implementation fails to hide the API key in these cases:

 - `https://base.gateway.tenderly.co/76zCIaEVWnjMIiV2gLWxKz`
 -  `https://eth.llamarpc.com/sk_llama_3b8dcfb486797e66de26feef7df7do55`,
 -  `https://arbitrum-mainnet.infura.io/v3/099fc55e0de9251d80b18d7c74caa7c7`,
 
Note: The API keys are fake but maintain the same format as the real ones.


dbeal-eth (approved):
honestly if it was me I  would just hide all parts after the hostname but I like this

PR #1198, opened 2024-07-12
feat(builder): Publish packages with IPFS ref

mjlescano:
fixes https://linear.app/usecannon/issue/CAN-401/allow-to-publish-using-ipfsqm-hash
fixes https://linear.app/usecannon/issue/CAN-402/migrate-ipfs-for-ipfs-format-on-cli-commands
fixes https://linear.app/usecannon/issue/CAN-400/packagereference-should-throw-an-error-when-given-ipfsqm-reference

Error message for deprecated ipfs format:
image


packages/builder/src/package.ts
```diff
@@ -37,7 +37,7 @@ export interface PackagePublishCall {
 export class PackageReference {
   static DEFAULT_TAG = 'latest';
   static DEFAULT_PRESET = 'main';
-  static PACKAGE_REGEX = /^(?<name>@?[a-z0-9][A-Za-z0-9-]{1,}[a-z0-9])(?::(?<version>[^@]+))?(@(?<preset>[^\s]+))?$/;
+  static PACKAGE_REGEX = /^(?<name>[a-z0-9][A-Za-z0-9-]{1,}[a-z0-9])(?::(?<version>[^@]+))?(@(?<preset>[^\s]+))?$/;
```
dbeal-eth, line 40: we should avoid breaking changes and remain backwards compatible probably, its ok to still have this in the regex

dbeal-eth (changes requested):
chagnes are good but we need to preserve old functionality aka `@ipfs:Qm...` format), even if we ultimately think its deprecated.

please assume this is the policy always unless said otherwise.

dbeal-eth:
also just to confirm these changes work everywhere (aka every cli command, every ui, etc? universal consistency would be great)

packages/cli/src/index.ts
```diff
@@ -417,30 +419,26 @@ applyCommandsConfig(program.command('publish'), commandsConfig.publish).action(a
...
+    if (!isRegistered) {
+      console.log();
+      console.log(
+        gray(
+          `Package "${fullPackageRef}" not yet registered, please use "cannon register" to register your package first.\nYou need enough gas on Ethereum Mainnet to register the package on Cannon Registry`
+        )
+      );
+      console.log();
 
-    if (!options.skipConfirm) {
```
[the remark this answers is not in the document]
FuzzB0t: Good catch this wasnt meant to be deleted

packages/cli/src/helpers.ts
```diff
@@ -492,3 +498,48 @@ export function checkAndNormalizePrivateKey(privateKey: string | viem.Hex | unde
 
   return normalizedPrivateKeys.join(',') as viem.Hex;
 }
+
+/**
+ *
+ * @param ref reference string, can be a package reference or ipfs url
+ * @returns Package Reference string
+ */
+export async function getPackageReference(ref: string) {
```
FuzzB0t, line 507: @dbeal-eth see here

packages/builder/src/package.ts
FuzzB0t: See here
https://github.com/usecannon/cannon/pull/1198/files#diff-ddd0df427184818e574abaf99eb65fa94549dc69884461cc1b32f5236f1dadccR507

dbeal-eth: Does it also get normalized in the web UI? Does it need to be normalized in the web UI? I don't actually know

FuzzB0t: Ive changed the code on the website that uses '@ipfs://' to use `ipfs://` and it since the website uses the builder `getUrl` function and thats been changed to normalize `@ipfs` urls to `ipfs://` it shouldnt need any more changes on the website.

FuzzB0t (approved):

PR #1201, opened 2024-07-13
feat(website): add selector tooltip

noahlitvin:
Maybe we right justify it with gray.400 small text so you can ctrl+f for them instead


packages/website/src/features/Packages/Function.tsx
```diff
@@ -577,7 +584,14 @@ export const Function: FC<{
                 alignItems="center"
                 gap={2}
               >
-                {toFunctionSignature(f)}
+                <Tooltip
+                  label={`Selector: ${toFunctionSelector(
```
dbeal-eth, line 588: does `toFunctionSignature` need ot happen before `toFunctionSelector`? I don't think its necessary, but please check.

dbeal-eth (approved):

mjlescano:
Closing as this is out of date

PR #1206, opened 2024-07-17
feat(cli): CLI UX improvements

saeta-eth:
Closes https://linear.app/usecannon/issue/CAN-335/cli-publish-ux-improvements

### DIFF
--- a/examples/router-architecture/cannonfile.toml
+++ b/examples/router-architecture/cannonfile.toml
@@ -22,6 +22,7 @@ contracts = [
   "Greeter",
 ]
 includeReceive = false
+highlight = true
 
 # Deploy a new Proxy using as implementation our just craeted Router
 [provision.proxy]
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -486,17 +486,34 @@ export async function build({
           ['Metadata', metaUrl],
         ])
       );
-      console.log(
-        bold(`Publish ${bold(fullPackageRef)} to the registry and pin the IPFS data to ${filteredSettings.publishIpfsUrl}`)
-      );
-      console.log(`> ${`cannon publish ${fullPackageRef} --chain-id ${chainId}`}`);
+
+      const isMainPreset = preset === 'main';
```
mjlescano, line 490: this should use `PackageReference.DEFAULT_PRESET`

packages/cli/src/commands/build.ts
```diff
@@ -486,17 +486,34 @@ export async function build({
...
+
+      if (isMainPreset) {
+        console.log(
+          bold(
+            `Publish ${bold(`${name}:${version}`)} to the registry and pin the IPFS data to ${
```
mjlescano, line 495: And for this combination it can use `PackageReference.packageRef` which doesn't include the preset.

mjlescano (approved):

PR #1215, opened 2024-07-21
fix(builder): Fix output for invalid fromCall.func

FuzzB0t:
When the `fromCall.func` function doesnt exist or is invalid, the error incorrectly prints out the value for the `func` defined function
image


dbeal-eth (approved):
can you share a picture of the fixed version also?

PR #1218, opened 2024-07-24
fix(builder): Fix output clash on var steps

FuzzB0t:
fixes https://linear.app/usecannon/issue/CAN-406/fix-output-clash-on-var-steps-with-description

https://uploads.linear.app/32382608-9e74-479c-9c61-9974bca22746/8e9b49db-3d3d-4314-bd3f-09ce0505e41b/afcd3a3d-42c3-4574-946e-2922444d20d7


packages/builder/src/definition.ts
```diff
@@ -104,7 +104,7 @@ export class ChainDefinition {
 
           for (const output of actionOutputs) {
             debug(`deps: ${fullActionName} provides ${output}`);
-            if (!this.dependencyFor.has(output)) {
+            if (!this.dependencyFor.has(output) || action == 'setting' || action == 'var') {
```
dbeal-eth, line 107: use triple equals `===`. surprised the linter didnt pick this one up.

dbeal-eth (changes requested):
i think this one definitely needs a test to verify this edge case doesn't reappear.

examples/sample-foundry-project/cannonfile.old.toml
```diff
@@ -3,15 +3,16 @@ version = "<%= package.version %>"
 description = "Simple project to verify the functionality of cannon"
 keywords = ["sample", "greeter"]
 
-[setting.value]
```
dbeal-eth, line 6: this file is specifically designed to test the behavior of the old `setting`s format and other things before those renaming changes, so changing from `setting` to `var` here is not a good idea

dbeal-eth (dismissed):
also still think we would benefit from specific, purpose built test, not in e2e, that verifies this capability on the `definition.ts` file.

packages/builder/src/definition.ts
```diff
@@ -104,7 +104,7 @@ export class ChainDefinition {
 
           for (const output of actionOutputs) {
             debug(`deps: ${fullActionName} provides ${output}`);
-            if (!this.dependencyFor.has(output)) {
+            if (!this.dependencyFor.has(output) || action === 'setting' || action === 'var') {
```
mjlescano, line 107: This wouldn't work weird when having a `var` that uses another `var`? Meaning that it wont be able to find correctly the dependency?

mjlescano (approved):

PR #1220, opened 2024-07-25
feat(website): add sentry

noahlitvin:


dbeal-eth (approved):
bunch of random sentry "example" files are also included. should those be removed?

noahlitvin:
> bunch of random sentry "example" files are also included. should those be removed?

The example page lets us trigger a test error to confirm things are working.

PR #1225, opened 2024-07-26
feat(cli): console wrapper

saeta-eth:
This PR introduces a wrapper around the console object to standardize message display and reduce the risk of inadvertently leaving console.log statements in the code.

Changes:
- Implemented a new log function to replace direct calls to console.log.
- Direct usage of console.log will now trigger a lint error.

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -177,7 +177,7 @@ export async function build({
 
   // Update pkgInfo (package.json) with information from existing package, if present
   if (oldDeployData) {
-    console.log(gray(`  ${fullPackageRef} (Chain ID: ${chainId}) found`));
+    log(gray(`  ${fullPackageRef} (Chain ID: ${chainId}) found`));
     if (!wipe) {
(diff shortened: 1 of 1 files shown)


FuzzB0t (approved):

dbeal-eth (approved):
interesting solution. as long as people dont get in the habit of just typing `info` when they want to debug log something it should be fine.

i think its worth mentioning we already had checks in place to ensure no log messages. However, in the cli project specifically, this check was completely disabled since the cli does so much logging anyway.

nicosampler (approved):
Awesome!

Recently, I've used this library https://github.com/pinojs/pino to wrap the logs, pretty similar to what you have done here. The advantage of Pino is that it uses a plugin system that allows you to easily manipulate how the logs are outputted. For example, it allows you to log to the console and, at the same time, for logs of type errors, write them to a log.errors file and/or to a cloud service.

The abstraction you have done here does almost all the work required to integrate Pino. If in the future we need more advanced logging functionality, we could simply add the Pino initialization here: packages/cli/src/util/console.ts and everything should keep working.

PR #1227, opened 2024-07-29
fix(website): hide loading when re-fetching txs information

nicosampler:
This PR adds a UI improvement when transactions are re-fetching. Now it does not show a spinner when a new re-fetch is triggered, for example, when re-visiting the tab. 

Closes #358.


dbeal-eth (approved):
are there any other places where we have made similar mistakes?

anyway lets merge this asap have heard tons of people complaining about this :laughing:

PR #1243, opened 2024-08-08
fix(builder): Force use of array on invoke target

FuzzB0t:
Turns out the error only happened because we allowed targets to be defined as strings as well which was causing the `getInputs` function to incorrectly return values, causing the computation of dependencies to fail.


mjlescano (approved):

dbeal-eth (approved):
We may want to double check the board, but I thought that previously we determined that we thought it was useful to have the ability to specify Target as non-array in the past. If that was the case, we may want to rethink our approach here and fix it so it is able to take non-array in the get inputs function. 

Otherwise it looks good

dbeal-eth (approved):

PR #1248, opened 2024-08-08
fix(website): cypress tests

saeta-eth:
This PR aims to fix the Cypress tests for the dapp. Recent changes to the Interact section have broken some tests.


nicosampler (approved):
Nice catch Seba!

PR #1249, opened 2024-08-09
etherscan contracts `auto-verify`

dbeal-eth:
unfortunately this is not fully working because etherscan does not support a particular API feature which would make this practical for supporting many chains for us, but I think we should still merge because the feature will likely be used soon and there are some ancillery changes that will likely be useful for later PRs


mjlescano (approved):

mjlescano:
Should this PR be marked as Draft for now?

mjlescano:
closing due to inactivity

PR #1250, opened 2024-08-09
fix(cli): Add better outputs for dry run and allow dry run on local network

FuzzB0t:
image

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -124,6 +124,7 @@ export async function build({
   }
 
   const chainId = await provider.getChainId();
+
   const chainInfo = getChainById(chainId);
   const chainName = chainInfo?.name || 'unknown chain';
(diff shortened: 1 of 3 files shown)


mjlescano:
There's a lint issue, but aside from that 🚀

mjlescano (approved):

dbeal-eth (dismissed):
need tests specifically addressing the resolved issues. for example not clear if this actually fixes the "disallow dry run flag on local build" fix. also in general, quite a lot of changes here and need to verify them

FuzzB0t:
> need tests specifically addressing the resolved issues. for example not clear if this actually fixes the "disallow dry run flag on local build" fix. also in general, quite a lot of changes here and need to verify them

yeah technically this is enabling dry runs on local builds so I will fix that and add tests.

mjlescano (approved):

PR #1253, opened 2024-08-12
fix(alter): update contract address on subpackages

mjlescano:
Closes https://linear.app/usecannon/issue/CAN-440/cannon-alter-packagename-set-contract-address-chain-id-xx-subpkg

This PR was used to debug the `set-contract-address` cannon command when editing subpackages.

1. ~~It adds the `--no-publish` flag to be able to test out new ipfs builds without having to publish to the registry~~
2. Rewrites the `set-contract-address` to make it more readable and make sure that only finds 1 step by the given name
3. Adds params validations in case of missing commands and some extra `debug` statements.

### DIFF
--- a/packages/cli/src/commands/alter.ts
+++ b/packages/cli/src/commands/alter.ts
@@ -227,19 +227,29 @@ export async function alter(
       }
 
       break;
-    case 'set-contract-address':
-      // find the steps that deploy contract
-      for (const actionStep in deployInfo.state) {
-        if (
-          deployInfo.state[actionStep].artifacts.contracts &&
-          deployInfo.state[actionStep].artifacts.contracts![targets[0]]
-        ) {
-          deployInfo.state[actionStep].artifacts.contracts![targets[0]].address = targets[1] as viem.Address;
-          deployInfo.state[actionStep].artifacts.contracts![targets[0]].deployTxnHash = '';
-        }
+    case 'set-contract-address': {
+      const [targetContractName, targetAddress] = targets;
+
+      if (!viem.isAddress(targetAddress)) {
+        throw new Error(`Invalid address given: "${targetAddress}"`);
       }
 
+      // find the step that deploy contract
+      const [targetStep] = Object.values(deployInfo.state).filter(
+        (step) => !!step.artifacts?.contracts?.[targetContractName]
+      );
+
+      if (!targetStep) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/commands/alter.ts
```diff
@@ -227,19 +228,27 @@ export async function alter(
...
+      // find the step that deploy contract
+      const [targetStep] = Object.values(deployInfo.state).filter((step) => !!step.artifacts?.contracts?.[targetStepName]);
+
+      if (!targetStep) {
+        throw new Error(`Could not find contract by step name "${targetStepName}"`);
       }
 
+      debug(`setting "${targetStepName}" address to "${targetAddress}"`);
+
+      targetStep.artifacts!.contracts![targetStepName].address = targetAddress;
```
dbeal-eth, line 247: looks like the var shouldn't be called `targetStepName`, and instead called `targetContractName`

packages/cli/src/commands/alter.ts
```diff
@@ -315,7 +324,9 @@ export async function alter(
     }
   }
 
-  await resolver.publish([fullPackageRef], chainId, subpkgUrl, metaUrl || '');
+  if (!noPublish) {
```
dbeal-eth, line 327: ooc why was `noPublish` option added to this command? is it getting in the way of something?

packages/cli/src/commandsConfig.ts
```diff
@@ -379,6 +379,10 @@ const commandsConfig = {
         flags: '-p --preset <preset>',
         description: '(DEPRECATED) Preset of the deployment to alter',
       },
+      {
+        flags: '--no-publish',
+        description: 'Do not publish the result to the regisgtry, just pin the new IPFS artifacts.',
```
dbeal-eth, line 384: it doesn't normally publish "to the registry", it pubishes to the local tags which can then be published

dbeal-eth (approved):
can we add tests

packages/cli/src/commands/alter.ts
mjlescano: I was needing it to be able to test it out more easily, without it trying to publish to the registry the result.

packages/cli/src/commandsConfig.ts
mjlescano: Ah perfect, thought that it was to the registry, will remove it then.

PR #1255, opened 2024-08-13
fix(website): publish utility was not loading package correctly

mjlescano:
This PR fixes a bug where the `PublishUtility` was trying to fetch the package ipfs url by ipfs url instead of package name. With this fix it just returns the given ipfs url and correctly resolves the `fullPackageRef` when necessary.

### DIFF
--- a/packages/website/src/hooks/cannon.ts
+++ b/packages/website/src/hooks/cannon.ts
@@ -388,21 +388,22 @@ export function useCannonPackage(packageRef?: string, chainId?: number) {
     enabled: !!pkgUrl,
   });
 
-  const fullPackageRef = ipfsQuery.data?.fullPackageRef;
-
   const registryQueryMeta = useQuery({
-    queryKey: ['cannon', 'registry-meta', fullPackageRef, packageChainId],
+    queryKey: ['cannon', 'registry-meta', packageRef, packageChainId],
     queryFn: async () => {
-      if (typeof fullPackageRef !== 'string' || fullPackageRef.length < 3) {
+      if (typeof packageRef !== 'string' || packageRef.length < 3) {
         return null;
       }
 
-      const metaUrl = await registry.getMetaUrl(fullPackageRef, packageChainId);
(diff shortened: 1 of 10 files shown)


packages/website/src/hooks/cannon.ts
```diff
@@ -331,23 +331,26 @@ export function useCannonPackage(packageRef: string, chainId?: number) {
...
-    queryKey: ['cannon', 'registry', packageRef, packageChainId],
+    queryKey: ['cannon', 'registry-url', packageRef, packageChainId],
     queryFn: async () => {
-      if (!packageRef || packageRef.length < 3) {
+      if (typeof packageRef !== 'string' || packageRef.length < 3) {
         return null;
       }
 
+      if (packageRef.startsWith('ipfs://')) return { url: packageRef };
+      if (packageRef.startsWith('@ipfs:')) return { url: packageRef.replace('@ipfs:', 'ipfs://') };
```
mjlescano, line 341: The main fix for this PR consists on just correctly handling when given a deploy url.

nicosampler (approved):

packages/website/src/hooks/cannon.ts
```diff
@@ -385,20 +388,40 @@ export function useCannonPackage(packageRef: string, chainId?: number) {
...
+
+      const metaUrl = await registry.getMetaUrl(fullPackageRef, packageChainId);
+
+      if (metaUrl) {
+        return { metaUrl };
+      } else {
+        throw new Error(`package not found: ${fullPackageRef} (${packageChainId})`);
+      }
+    },
+    refetchOnWindowFocus: false,
```
nicosampler, line 408: does it make sense to disable it globally when Tanstak/Query is initialized.

I've seen this refech in many places and it is quite odd. I think the default should be disabled and enabled when required on specific quieres.

mjlescano: Yup, I think you are right, best would be to disable it globally. We only really need it on the signatures list, when signing a transaction.

PR #1260, opened 2024-08-14
fix(builder): throw an error when simulating create2 deployments

saeta-eth:


packages/builder/src/steps/deploy.ts
```diff
@@ -344,7 +344,7 @@ const deploySpec = {
               _.assign(txn, overrides, { account: signer.wallet.account || signer.address })
             );
 
-            deployAddress = viem.zeroAddress;
+            throw new Error("Simulation failed. We can't decode the revert message.");
```
dbeal-eth, line 347: can we have a better error here? like `The CREATE2 contract seems to be failing in the constructor. However, we were not able to get a stack trace`

dbeal-eth (approved):

PR #1261, opened 2024-08-14
`cannon diff` command

dbeal-eth:
this command allows for comparing aspects of cannon state with things.

for the time being, this PR focuses on comparing the cannon state contracts with the ones in a foundry project, highlighting any differences.

notes:
* its possible to filter contracts to compare and source files to compare using the provided options
* exit code matches the nubmer of "differences" found, so only successful output is no diffs (usually what people are checkign for)
* we could do something very similar to hardhat, but lets wait to see how the current impl works with stakeholders


packages/cli/src/commands/diff.ts
```diff
@@ -0,0 +1,171 @@
+import Debug from 'debug';
+import * as Diff from 'diff';
+import * as viem from 'viem';
+import { getFoundryArtifact, buildContracts } from '../foundry';
+import { bold, yellow } from 'chalk';
+import { ChainDefinition, ChainBuilderRuntime, DeploymentInfo, getArtifacts } from '@usecannon/builder';
+import { PackageReference } from '@usecannon/builder/dist/src/package';
```
mjlescano, line 7: `PackageReference` is exported by the index:
```suggestion
import { ChainDefinition, ChainBuilderRuntime, DeploymentInfo, getArtifacts, PackageReference } from '@usecannon/builder';
```

mjlescano (approved):

FuzzB0t (approved):

PR #1287, opened 2024-08-16
chore(website): improve cypress performance

saeta-eth:
This PR aims to improve the performance of the Cypress test suite. The goal is to replace the Next.js development server with the output of the build process and run a static file server instead. This approach makes testing faster and more reliable than before. I've included some screenshots showing the improvements:

Before
Screenshot 2024-08-17 at 1 16 23 PM

After:
Screenshot 2024-08-17 at 1 15 53 PM

Overall, the Cypress tests finish in approximately ~8 minutes, compared to the previous ~20 minutes:

Before:

Screenshot 2024-08-17 at 1 24 31 PM

After:

Screenshot 2024-08-17 at 1 25 20 PM

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -2,6 +2,7 @@ import {
   build as cannonBuild,
   CANNON_CHAIN_ID,
   CannonRegistry,
+  CannonSigner,
   ChainArtifacts,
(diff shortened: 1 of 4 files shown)


nicosampler (approved):
amazing! 🥳

PR #1291, opened 2024-08-18
feat(builder): change behavior of create2 to fail on existing contract

dbeal-eth:
also do some refactoring wrt handling of `PackageReference` with `ChainDefinition`, which lead to a bit of a rabbit hole

fixes https://linear.app/usecannon/issue/CAN-136/change-behaviour-of-contract-cannon-build-with-create2-to-fail-on


packages/builder/src/steps/clone.ts
```diff
@@ -61,7 +61,7 @@ const cloneSpec = {
     }
 
     config.sourcePreset = template(config.sourcePreset)(ctx);
-    config.targetPreset = template(config.targetPreset)(ctx) || `with-${packageState.name}`;
+    config.targetPreset = template(config.targetPreset)(ctx);
```
mjlescano, line 64: Wouldn't this change make the subpackages be published as `main` by default?

mjlescano (approved):

packages/builder/src/steps/clone.ts
mjlescano: Here's the same change, but also changed on the `configInject()` fn, which on should go? https://github.com/usecannon/cannon/pull/1292/files#diff-950232bdec873b2ee3a9f82132b20b978b8ae9fcf3be3a475e613efee95046f6R64w

dbeal-eth: config inject is not supposed to materially alter the , only perform interpolation and type extraction and formatting of data as needed basically

setting `with-${pkgName}` there was kind of an edge case because it was formatting default data, but it gets in the way nad has wierd side effects in the . So i set it up this time to not set the default here and instead it gets set in the `exec` function

not sure what your second comment is trying to say.

dbeal-eth:
force merging for now because the e2e test failure is not reproducible on my machin ewhen running the same command with build and all

PR #1293, opened 2024-08-19
fix: Invalid import references

noisekit:
Because code of `cli` references sources of `builder` instead of referencing compiled dist files tsc includes whole cannon source when compiling downstream projects and fails to build.

## Before
image

## After
image


noahlitvin (approved):

mjlescano (approved):

mjlescano:
This is going to be added to `main` here: https://github.com/usecannon/cannon/pull/1293

PR #1305, opened 2024-08-21
fix(website): Use Ipfs reference for partial deploy on web 

FuzzB0t:
fixes https://linear.app/usecannon/issue/CAN-444/add-deployments-from-partial-deploy-file

This ticket redesigns the deploy page and removes the need to specify cannonfile url in partial deployments on the web app.

Below is an example of running a partial deployment (I ran the same thing on the current instance of the website and confirmed the error response was the same on there)

Without cannonfile ref:

https://www.loom.com/share/054e6dda24aa47aebd4f0c729ed83f61

With cannonfile ref:

https://www.loom.com/share/aa0414806f364c679e8b9978525cb5aa

Also added the ability to specify chain id on pin command since it was not pinning partial deployment hashes on other networks


packages/cli/src/commandsConfig.ts
```diff
@@ -414,7 +414,13 @@ const commandsConfig = {
         description: 'IPFS hash to write deployment data for',
       },
     ],
-    options: [...debugVerbosity],
+    options: [
```
FuzzB0t, line 417: Added chain id option to pin command because the pin command fails when trying to pin a package that was built on a live network

mjlescano (changes requested):
Currently the UI is allowing you to just paster a cannonfile without a partial deployment file, which it shouldn't:
Screenshot 2024-08-26 at 18 54 22

mjlescano:
When landing for the first time on the gitops deployment page, directly from the link or navigating, there's no radio button selected by default, when the first one should be selected:
Screenshot 2024-08-27 at 11 46 51

mjlescano (approved):

PR #1307, opened 2024-08-22
chore(website): try to consume API types from CI

nicosampler:
DO NOT MERGE


packages/website/src/hooks/api/usePackage.ts
```diff
@@ -1,10 +1,54 @@
...
+  }
+};
+
+const fetchPackageByName = async ({ queryKey }: { queryKey: string[] }) => {
+  const [, name] = queryKey;
+  try {
+    const response = await axios.get<{ total: any; data: ApiDocument[] }>(`packages/${name}`, {
+      baseURL: externalLinks.API_CANNON,
+    });
+    return response.data;
```
mjlescano, line 24: Shouldn't this be `response.data.data` ?
Not sure about the `response` object, but the API on `GET /packages/:packageName` responds with:
```json
{
  status: number,
  total: number,
  data: { ... }
}
```

mjlescano (approved):

packages/website/src/hooks/api/usePackage.ts
mjlescano: Ah perfect, just wanted to make sure that it is being used correctly on the Components

PR #1320, opened 2024-08-27
fix(cli): Refactor pin command

FuzzB0t:
This PR refactors the pin command to have its own function in the CLI, however to avoid duplicating code I think it might be better if the pin command stays within a function in the builder (named `pin` respectively) so that both the cli's `pin` command and the builder's `preparePublishPackage` function can both use the code.

Might be nice to add the ability to pin from the webapp and add a button for explicit navigation to `/ipfs` which will let you download and/or pin hashes from/to IPFS

fixes https://linear.app/usecannon/issue/CAN-467/cannon-pin-should-not-require-chain-id-when-pinning-using-an-ipfs-hash


mjlescano (approved):

dbeal-eth:
@FuzzB0t is this done? are we goingt o fix the test failures?

FuzzB0t:
> @FuzzB0t is this done? are we goingt o fix the test failures?

Yes its done, the tests are failing due to another intermittent error that has been happening on CI for a while.
Merging it now.

PR #1329, opened 2024-08-30
add snaxchain temp

dbeal-eth:


noahlitvin (changes requested):
Better to put in this file? https://github.com/usecannon/cannon/blob/main/packages/cli/src/chains.ts

PR #1344, opened 2024-09-03
chore: v2.16.8-alpha.0

FuzzB0t:


mjlescano (changes requested):
This is not going to be merged as a new stable release is going out: https://github.com/usecannon/cannon/pull/1347

PR #1346, opened 2024-09-03
feat(website): Deploy contracts through website on web deployer

dbeal-eth:
fixes https://linear.app/usecannon/issue/CAN-432/cannon-deployer-service


dbeal-eth:
this has been merged into automatic versioning because they affect a lot of the same code.

PR #1351, opened 2024-09-03
fix(builder): Replace get calls to registry with axios

FuzzB0t:
Replaces `getUrl` and `getMetaUrl` to use axios to make rpc calls to the registry

### DIFF
--- a/packages/builder/src/registry.ts
+++ b/packages/builder/src/registry.ts
@@ -9,6 +9,7 @@ import CannonRegistryAbi from './abis/CannonRegistry';
 import { prepareMulticall, TxData } from './multicall';
 import { PackageReference } from './package-reference';
 import { CannonSigner } from './types';
+import axios from 'axios';
 
 const debug = Debug('cannon:builder:registry');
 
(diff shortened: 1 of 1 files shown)


FuzzB0t:
> What is the motivation for this PR? I am out of the loop and it seems confusing

@saeta-eth 
We we're seeing a lot of issues with the registry get calls and RPC's so db mentioned we could optimize the calls by using axios to make rpc requests instead of viem. There may be other reasons for this PR but that was the main one AFAIK.

packages/builder/src/registry.ts
```diff
@@ -480,7 +480,36 @@ export class OnChainRegistry extends CannonRegistry {
...
+      },
+      {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+    );
+
+    if (response.data && response.data.error) {
+      throw response.data.error.message;
```
mjlescano, line 503: Is the `message` prop a string? in that case it could be thrown as an error to have a full stack trace:
```suggestion
      throw new Error(response.data.error.message);
```

packages/builder/src/registry.ts
```diff
@@ -480,7 +480,36 @@ export class OnChainRegistry extends CannonRegistry {
       ],
     });
 
-    return (url as string) || null;
+    const response = await axios.post(
```
mjlescano, line 483: This call could be moved to a helper like `contractFunctionCall` so this method does not get super long.

PR #1352, opened 2024-09-04
fix(inspect): debug info

mjlescano:
closes https://linear.app/usecannon/issue/CAN-457/cannon-inspect-has-an-extraneous-log

This PR makes sure that the build is not generating any extraneous logs when running `cannon inspect --json`, as the only expected output should be the JSON result.

Nevertheless, debug logs can still be used with:
```
DEBUG=cannon:cli:inspect dcannon inspect ...
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "lint:fix:sol": "prettier --write '{packages,examples}/*/{contracts,src}/**/*.sol'",
     "lint:fix": "pnpm run '/^lint:fix:(js|sol)/'",
     "build": "pnpm -r --filter @usecannon/builder --filter @usecannon/cli --filter hardhat-cannon --filter @usecannon/api run build",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -10,6 +10,7 @@
     "lint:fix:sol": "prettier --write '{packages,examples}/*/{contracts,src}/**/*.sol'",
     "lint:fix": "pnpm run '/^lint:fix:(js|sol)/'",
     "build": "pnpm -r --filter @usecannon/builder --filter @usecannon/cli --filter hardhat-cannon --filter @usecannon/api run build",
+    "watch": "pnpm -r --parallel --filter @usecannon/builder --filter @usecannon/cli run watch",
```
mjlescano, line 13: added a helper script to be able to just run `pnpm watch` on the root, instead of having to open several tabs

PR #1357, opened 2024-09-05
feat(website): CAN-474 link to navigate to a package.

nicosampler:
navigate to a package using a <a> link.

Closes  #CAN-474


noahlitvin (changes requested):
Can we make the entire row a link? Like `as={Link}` on the the Tr probably?

dbeal-eth (commented):
Need background as to why we are doing this.

Also I want to confirm, this is still static-site compatible, right?

mjlescano (approved):

noahlitvin (approved):

PR #1358, opened 2024-09-06
feat(cli): Support `--anvil.*` flags

saeta-eth:
This PR makes a clearer distinction between Cannon and Anvil flags by prefixing Anvil flags with --anvil.*, improving clarity and making their origin explicit. It also includes the upgrade to the latest version of Commander.

Closes https://linear.app/usecannon/issue/CAN-508/add-foundry-params-to-all-cannon-commands-with-forge-and-anvil


dbeal-eth (approved):
would be really cool if we could progmatically just pass any arguments based on if they start with `--anvil.` prefix rather than having to code them here (that way if they add/change arguments in the future, it always works with user's setup)

mjlescano (approved):

PR #1362, opened 2024-09-07
fix(website): CAN-493 switch chain instead of sign

nicosampler:
Fixes the case where the button should show the text "Switch chain" instead of "Sign"


mjlescano (approved):

dbeal-eth (commented):
I just want to confirm but I thought previously we set it up so that if you click the sign button, it does both the switch Network and sign at the same time, so I'd rather use that functionality if possible. Do you have a video?

dbeal-eth (approved):

PR #1372, opened 2024-09-12
feat(create2): allow to set ifExists = "continue"

mjlescano:
closes https://linear.app/usecannon/issue/CAN-518/add-option-to-create-a-non-strict-create2-deploy-step


FuzzB0t (approved):

dbeal-eth (changes requested):
One potential problem with this setup is that it doesn't allow for the simultaneous setting of the create2 contract address and the no strict option. It makes me feel like we need to have a separate arg all together

packages/builder/src/schemas.ts
```diff
@@ -80,20 +80,29 @@ export const deploySchema = z
...
         /**
-         *    Contract deployer address.
-         *    Must match the ethereum address format
+         * Determines whether to deploy the contract using create2
+         */
+        ifExists: z
+          .enum(['continue'])
+          .optional()
+          .describe(
+            'Determines whether when trying to deploy the contract using CREATE2 it should continue if the target address already exists. Disabled by default'
```
dbeal-eth, line 101: 'When deploying a contract with CREATE2, determines the behavior when the target contract is already deployed (ex. due to same bytecode and salt). Set to `continue` to allow the build to continue if the contract is found to have already been deployed. By default, an error is thrown and the action is halted.'

packages/builder/src/steps/deploy.ts
```diff
@@ -278,7 +278,7 @@ const deploySpec = {
       if (config.create2) {
         const arachnidDeployerAddress = await ensureArachnidCreate2Exists(
           runtime,
-          typeof config.create2 === 'string' ? (config.create2 as viem.Address) : ARACHNID_DEFAULT_DEPLOY_ADDR
```
dbeal-eth, line 281: i think the previous condition here was better

dbeal-eth (approved):
more descriptive comment

PR #1378, opened 2024-09-14
fix(website): gitops deployment was missing chainId

mjlescano:
This PR fixes the gitops deployment page, by doing the actual fix first, and also fixing the types on the builder so this kind of error is less likely.
1. The website hook was initialising the simulated build without chainId, that's fixed here: https://github.com/usecannon/cannon/pull/1378/files#diff-1c16184636ccf287d09107e881baa71fba0cc29067ff908c276919cf5922d6b4R266
2. Now, when fetching deployment packages we make sure it includes `chainId` to be consistent with old deployments, kinda like a migration: https://github.com/usecannon/cannon/pull/1378/commits/9c6c542b628e755ae4ac5e7f1c89df834da64a64


packages/builder/src/runtime.ts
```diff
@@ -88,6 +88,12 @@ export class CannonStorage extends EventEmitter {
 
     const deployInfo: DeploymentInfo = await this.readBlob(uri);
 
+    // Make sure the deployment artifacts include the chainId, this is necessary
+    // because old deployments already published didn't include it.
+    if (!Number.isSafeInteger(deployInfo.chainId)) {
```
dbeal-eth, line 93: is this condition necessary? can we just always set it?

packages/builder/src/types.ts
```diff
@@ -254,7 +254,7 @@ export type DeploymentInfo = {
   miscUrl: string;
 
   // EVM chain which this deployment is for
-  chainId?: number;
+  chainId: number;
```
dbeal-eth: this type should remain optional

packages/website/src/hooks/cannon.ts
```diff
@@ -76,6 +76,7 @@ export function useCannonBuild(safe: SafeDefinition | null, def?: ChainDefinitio
   const { addLog } = useLogs();
   const settings = useStore((s) => s.settings);
 
+  const [finishedBuilding, setFinishedBuilding] = useState(false);
   const [isBuilding, setIsBuilding] = useState(false);
```
dbeal-eth, line 80: lets collapse `isBuilding` and `finishedBuilding` into a single state `status` which is a string that can be of value `building`|`built`|`idle` or whatever

the booleans can still exist, but they should be computed from this one state

dbeal-eth (changes requested):

packages/website/src/hooks/cannon.ts
```diff
@@ -76,8 +76,8 @@ export function useCannonBuild(safe: SafeDefinition | null, def?: ChainDefinitio
   const { addLog } = useLogs();
   const settings = useStore((s) => s.settings);
 
-  const [isBuilding, setIsBuilding] = useState(false);
-  const [buildStatus, setBuildStatus] = useState('');
+  const [buildMessage, setBuildMessage] = useState('');
+  const [buildStatus, setBuildStatus] = useState<'idle' | 'building' | 'success' | 'error'>('idle');
```
dbeal-eth, line 80: :clap:

dbeal-eth (approved):

PR #1383, opened 2024-09-17
fix(builder): ctx timestamp type

mjlescano:
Closes https://linear.app/usecannon/issue/CAN-538/on-cannonfiles-the-timestamp-variable-should-be-of-type-number

This PR fixes the necessity on having to cast the `timestamp` variable to `number` on cannonfiles on cases like this one:
Screenshot_2024-09-16_at_3 42 59_PM


dbeal-eth (approved):
seems ok

PR #1408, opened 2024-09-23
fix(publish): chain id setup

mjlescano:
This PR fixes two bugs:
1. When running `cannon publish` including `clone` steps, it was wrongly publishing the subpackages with the `13370` chain instead of the target chain, that is addressed here: https://github.com/usecannon/cannon/pull/1408/files#diff-756d4c0d0bd036727ea6922554c95272acc0b2e94bd8311308f1a69574d94f82R98
2. When running `cannon build`, the imports of clone steps were having always the chainId `13370`, this is fixed here: https://github.com/usecannon/cannon/pull/1408/files#diff-950232bdec873b2ee3a9f82132b20b978b8ae9fcf3be3a475e613efee95046f6R256


FuzzB0t (approved):

.nx/workspace-data/d/server-process.json
```diff
@@ -0,0 +1 @@
+{"processId":63247}
```
dbeal-eth, line 1: remove

.nx/workspace-data/lockfile.hash
```diff
@@ -1 +1 @@
-15584140240147292807
\ No newline at end of file
+11580160284360395692
```
dbeal-eth, line 1: remove?

packages/builder/src/steps/clone.ts
```diff
@@ -253,7 +253,7 @@ const cloneSpec = {
       state: builtState,
       meta: deployInfo.meta,
       status: partialDeploy ? 'partial' : 'complete',
-      chainId,
+      chainId: runtime.chainId,
```
dbeal-eth, line 256: this is wrong. `clone` steps should be pulling from chain ID 13370 by default. This is as intended. Please undo.

dbeal-eth (changes requested):

dbeal-eth:
fix confirmed, going to merge this as is.

dbeal-eth (approved):

PR #1410, opened 2024-09-24
fix(website): CAN-466 safeInputAddress 

nicosampler:
Fixes the safeInputAddress interaction.


mjlescano:
It appears that is breaking:
1. The `X` that is used for unselecting the currently selected item
2. The deletion from the dropdown of the currently selected item

https://github.com/user-attachments/assets/2bff41f6-d7b8-41dd-a015-0083d7d51bb5

mjlescano (changes requested):
It is throwing an error when arriving to a custom chain url, e.g.:

http://localhost:3000/deploy?chainId=2192&address=0xF5d66dC9B6963c61D7dacB1208145397cAc2e678

Screenshot 2024-10-04 at 10 10 36

and it is working on production with the same url:
image

mjlescano (approved):

mjlescano:
Moving to approved, as I was missing the custom chain rpc setting.

PR #1413, opened 2024-09-24
fix(build): remove NX

mjlescano:
Remove [NX Cloud](https://nx.dev/nx-cloud) as with pnpm is not that necessary anymore.


dbeal-eth (approved):
:+1:

PR #1417, opened 2024-09-27
fix(builder): remove constructor from router abi

saeta-eth:
Closes https://linear.app/usecannon/issue/CAN-542/router-abi-should-not-include-constructor-items

### DIFF
--- a/packages/builder/src/util.ts
+++ b/packages/builder/src/util.ts
@@ -101,7 +101,7 @@ export function getMergedAbiFromContractPaths(ctx: ChainArtifacts, paths: string
     });
 }
 
-export const stripConstructorFromAbi = (abi: any[]) => {
+export const removeConstructorFromAbi = (abi: any[]) => {
   return abi.filter((a) => a.type !== 'constructor');
 };
 


mjlescano (approved):

packages/builder/src/steps/router.ts
```diff
@@ -135,13 +140,16 @@ const routerStep = {
     const inputData = getCompileInput(contractName, sourceCode, evmVersion);
     const solidityInfo = await compileContract(contractName, sourceCode, evmVersion);
 
-    // the abi is entirely basedon the fallback call so we have to generate ABI here
+    // the ABI is entirely based on the fallback call so we have to generate ABI here
     const routableAbi = getMergedAbiFromContractPaths(ctx, config.contracts);
```
dbeal-eth, line 144: technically this variable name is now inaccurate because the constructor is not "routable"

dbeal-eth (commented):
technically we should probably also remove constructor from abi in the `invoke` `abiOf` field also, which uses the same function

PR #1419, opened 2024-09-30
feat(website): add snaxchain color

noahlitvin:


noahlitvin:
How are these types generated?

PR #1433, opened 2024-10-01
fix(cli): update decode/trace documentation and logging

mjlescano:


packages/cli/src/commands/config/index.ts
```diff
@@ -617,7 +617,8 @@ export const commandsConfig: CommandsConfig = {
     ],
   },
   trace: {
-    description: 'Get a full stack trace for a transaction hash or explicit transaction call',
+    description:
+      'Get a full stack trace for a transaction hash or explicit transaction call. This command will sping a fork and simulate the given transaction.',
```
dbeal-eth: typo after "This command will spin up a fork"...

packages/cli/src/commands/config/index.ts
```diff
@@ -669,15 +670,16 @@ export const commandsConfig: CommandsConfig = {
     ],
   },
   decode: {
-    description: 'decode transaction data using the ABIs of the given Cannon package',
+    description:
+      'Decode the given data using the ABIs from the specified Cannon package. It will try to parse the data as Function Call, Error or Event. If a tx hash is given it will fetch it using the rpc and try to parse the input.',
```
dbeal-eth, line 674: I'm pretty sure we don't support txn hashes on this... unless this feature was added recently and I didnt realize?

packages/cli/src/commands/decode.ts
```diff
@@ -57,16 +57,25 @@ export async function decode({
...
 
-    const transaction = await provider.getTransaction({
+    const tx = await provider.getTransaction({
       hash: data,
     });
 
-    inputData = transaction.input;
+    inputData = tx.input;
+
+    log('     RPC: ', provider.transport.url);
```
dbeal-eth, line 66: do we need to clean/strip the url output, in case its sensitive?

packages/cli/src/commands/decode.ts
```diff
@@ -57,16 +57,25 @@ export async function decode({
...
+    log('Chain ID: ', tx.chainId);
+    log(' TX hash: ', tx.hash);
+    if (tx.from) log('    From: ', tx.from);
```
dbeal-eth, line 69: if from is not specified its probably best to say `From: undefined` rather than just not showing it.

packages/cli/src/commands/decode.ts
```diff
@@ -75,7 +84,7 @@ export async function decode({
       log(errorMessage);
       return;
     }
-    throw new Error('Could not decode transaction data');
+    throw new Error('Could not decode transaction data with the given Cannon package');
```
dbeal-eth, line 87: add some advice like "Please confirm that the given data exists in the ABIs of the given package."

PR #1435, opened 2024-10-02
fix foundry build issues

dbeal-eth:
also significanlty improves the speed by removing calls to foundry commands. Warp speed.

also fix number of constructor arguments not being checked for contract deployment (tbh surprised viem doesn't do this automatically but whateves)


mjlescano (approved):
There are some failing tests that need to be fixed before merging

mjlescano:
Closing in favor of https://github.com/usecannon/cannon/pull/1439

PR #1444, opened 2024-10-04
fix(website): CAN-540 safe tx service input.

nicosampler:
Fix: remove trailing slash if present when a safe tx service url is entered.
Feat: add URL validation  in safe tx service url input.
Refactor: Move Safe tx URL to a new file.


mjlescano (approved):
nice refactor!

PR #1458, opened 2024-10-06
feat(cli): forge build for tests by default

noahlitvin:


dbeal-eth (changes requested):
Please also add `--skip-buils` which suppresses the running of the `doBuild` function

packages/cli/src/foundry.ts
```diff
@@ -153,3 +161,40 @@ export async function getFoundryArtifact(name: string, baseDir = '', includeSour
...
+    '--ast',
+    ...fromFoundryOptionsToArgs(pickForgeBuildOptions(options), forgeBuildOptions),
+  ];
+
+  const forgeBuildProcess = spawn('forge', forgeBuildArgs, { cwd: projectDirectory, shell: true });
+
+  await new Promise((resolve, reject) => {
+    forgeBuildProcess.on('exit', (code) => {
+      if (code === 0) {
+        log(gray('forge build succeeded'));
```
juanpernu, line 187: And here, maybe add something like: `Forge build succeeded, if you want to skip the build use --no-build flag.`

PR #1465, opened 2024-10-09
fix(builder): correctly parse accesses on config.target for invoke steps

mjlescano:


packages/builder/src/definition.ts
```diff
@@ -344,6 +344,7 @@ export class ChainDefinition {
           possibleFields.push(baseName);
         }
       }
+
```
dbeal-eth, line 347: unnecessary whitespace add

dbeal-eth (approved):
nice.

PR #1468, opened 2024-10-10
fix: dry run gas estimate

FuzzB0t:
Now adds a warning on dry run when users wallet balance is below the cost of the deploy
image

fixes https://linear.app/usecannon/issue/CAN-558/show-how-much-gas-is-necessary-after-a-build-with-dry-run


dbeal-eth (changes requested):
hmm... this solution seems unusual. wouldn't it be better to record the balance within the `build` function body rather than in the `getSigner` function? the `getDefaultSigner` function will return the signer which can then be `getBalance` called

FuzzB0t:
> hmm... this solution seems unusual. wouldn't it be better to record the balance within the `build` function body rather than in the `getSigner` function? the `getDefaultSigner` function will return the signer which can then be `getBalance` called

I tried calling `getBalance` using the default signer in the `build` function before we set the balance on the getSigner function but it kept returning the balance as `10000`.

dbeal-eth:
> > hmm... this solution seems unusual. wouldn't it be better to record the balance within the `build` function body rather than in the `getSigner` function? the `getDefaultSigner` function will return the signer which can then be `getBalance` called
> 
> I tried calling `getBalance` using the default signer in the `build` function before we set the balance on the getSigner function but it kept returning the balance as `10000`.

oh, that is because in the simulation, the impersonated account's addresses (such as the default deployer) will automatically be set very high to allow the release to go forward

seems like we should be finding some way to check the balance on the underlying provider connection directly?

PR #1469, opened 2024-10-10
fix(website): CAN-582

nicosampler:
Only show the execute button on queue deploy if there are no safe steps.

image

Tested with: https://github.com/nicosampler/weth/blob/main/cannonfile.nodeployers.toml

### DIFF
--- a/packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
+++ b/packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
@@ -534,6 +534,16 @@ export default function QueueFromGitOps() {
   const hasDeployers = Boolean(
     cannonDefInfo.def?.getDeployers()?.length ?? 0 > 0
   );
+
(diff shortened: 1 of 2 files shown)


packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -1027,7 +1027,8 @@ export default function QueueFromGitOps() {
             <Box>
               <NoncePicker safe={currentSafe} handleChange={setPickedNonce} />
               <HStack gap="6">
-                {stager.execConditionFailed ? (
+                {stager.execConditionFailed &&
+                (buildState.result?.safeSteps.length || 0) > 0 ? (
```
mjlescano, line 1031: Perfect, this check should also be added on the "Execute Outside Safe Txns", because if not it would allow the user to deploy those contracts but not add them to any package.

mjlescano (approved):

PR #1473, opened 2024-10-11
feat(cli): forge build for tests by default

noahlitvin:


packages/cli/src/foundry.ts
```diff
@@ -153,3 +161,40 @@ export async function getFoundryArtifact(name: string, baseDir = '', includeSour
...
+        log(red('forge build failed'));
+        log(red('Make sure "forge build" runs successfully or use the --skip-compile flag.'));
+        return reject(new Error(`forge build failed with exit code "${code}"`));
+      }
+
+      resolve(null);
+    });
+  });
+} else {
+  log(yellow('Skipping forge build...'));
```
juanpernu, line 198: Here could be something like: `log(yellow('--no-build flag detected, skipping forge build...'));`?🤔

packages/cli/src/foundry.ts
```diff
@@ -153,3 +161,40 @@ export async function getFoundryArtifact(name: string, baseDir = '', includeSour
...
   };
 }
+
+
+export async function forgeBuild(options: Record<string, any>, cannonfile: string) {
+
+  const cannonfilePath = path.resolve(cannonfile);
+  const projectDirectory = path.dirname(cannonfilePath);
+
+log(bold('Building the foundry project...'));
```
juanpernu, line 171: This could be inside line `172`

mjlescano:
closing due to inactivity

PR #1474, opened 2024-10-15
feat(dev): docker-compose for dev environment

mjlescano:


packages/cli/src/util/on-keypress.ts
```diff
@@ -20,7 +20,9 @@ export default function onKeypress(handleKeyPress: (evt: KeyboardEvent, controls
     });
 
     readline.emitKeypressEvents(process.stdin, rl);
-    process.stdin.setRawMode(true);
+    if (process.stdin.isTTY) {
+      process.stdin.setRawMode(true);
+    }
```
mjlescano, line 25: This fixes a bug where the `cannon run` command is not working on non tty terminals, e.g. CI or `child_process.spawn`

mjlescano:
`supersim` cannot be used on Docker in MacOS at the moment, but they confirmed that there's a fix coming for it here: https://github.com/ethereum-optimism/supersim/issues/250

PR #1477, opened 2024-10-16
feat(cli): --write-script-format has foundry option

saeta-eth:
Closes https://linear.app/usecannon/issue/CAN-586/add-foundry-script-output-format-to-the-write-script-param

Example output:
```bash
#!/bin/bash

# Set your environment variables
export PRIVATE_KEY=<your_private_key>
export RPC_URL=<your_rpc_url>

# Start processing transactions
echo "Executing: [deploy.Greeter]"
cast send --private-key $PRIVATE_KEY --rpc-url $RPC_URL --create 0x...

echo "Executing: [invoke.setGreeting]"
cast send --private-key $PRIVATE_KEY --rpc-url $RPC_URL 0x...

echo "Transactions complete."

```

### DIFF
--- a/packages/cli/src/write-script/render-foundry.ts
+++ b/packages/cli/src/write-script/render-foundry.ts
@@ -10,8 +10,9 @@ import { Transform } from 'node:stream';
 const header = `#!/bin/bash
 
 # Set your environment variables
-export PRIVATE_KEY=<your_private_key>
-export RPC_URL=<your_rpc_url>
+# Set your environment variables
+# export PRIVATE_KEY=<your_private_key>
+# export RPC_URL=<your_rpc_url>
 
 # Start processing transactions
 `;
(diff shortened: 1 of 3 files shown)


packages/cli/src/write-script/render-foundry.ts
```diff
@@ -0,0 +1,74 @@
...
+ * This script is used to deploy contracts using Foundry Cast.
+ * It outputs a bash script that can be used to deploy contracts and execute transactions.
+ * Note: Make sure you add `.sh` extension to the output file.
+ */
+
+const header = `#!/bin/bash
+
+# Set your environment variables
+export PRIVATE_KEY=<your_private_key>
+export RPC_URL=<your_rpc_url>
```
mjlescano, line 14: This 2 lines should have working values so the script does not need to be modified to work, e.g. it can be easily generated during CI and just add the 2 env vars:
```suggestion
export PRIVATE_KEY=${PRIVATE_KEY:-""}
export RPC_URL=${RPC_URL:-""}
```

mjlescano (approved):

PR #1481, opened 2024-10-17
feat(router): CAN-596 add web-solc to compile on browser

mjlescano:
Closes https://linear.app/usecannon/issue/CAN-596/allow-to-build-the-router-on-website-using-solc

### DIFF
--- a/packages/builder/src/steps/router.ts
+++ b/packages/builder/src/steps/router.ts
@@ -111,6 +111,8 @@ const routerStep = {
         throw new Error(`contract not found: ${n}`);
       }
 
+      const contractName = n.replace('.', '_'); // Use step name, and replace '.' in case is pointing to an import.
+
       return {
         constructorArgs: contract.constructorArgs,
         abi: contract.abi,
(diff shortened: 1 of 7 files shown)


packages/builder/src/steps/router.ts
```diff
@@ -111,16 +111,18 @@ const routerStep = {
         throw new Error(`contract not found: ${n}`);
       }
 
+      const contractName = n.replace('.', '_'); // Use step name, and replace '.' in case is pointing to an import.
```
mjlescano, line 114: This fixes a bug where it was not possible to render a Router that included contracts from an imported package 🐛

packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -562,8 +554,7 @@ export default function QueueFromGitOps() {
     (onChainPrevPkgQuery.isFetched &&
       !prevDeployLocation &&
       tomlRequiresPrevPackage &&
-      !previousPackageInput) ||
-    canTomlBeDeployedUsingWebsite;
+      !previousPackageInput);
```
mjlescano, line 557: 🐛 This change fixes a bug were the "Preview Transactions Button" was disabled if the package was new.

PR #1496, opened 2024-10-22
chore(builder): switch to Viem's setCode for Ganache

saeta-eth:
This PR aims to remove a workaround that was implemented when Viem did not yet include the setCode function for Ganache. Related PR: https://github.com/wevm/viem/pull/2636


mjlescano (approved):
❤️ ❤️ ❤️

PR #1507, opened 2024-10-25
wip: show the list of txs and status of the txs that has to be executed outside the Safe. 

nicosampler:
This PR has two intentions:

1. Do a refactor over `useDeployerWallet`. Replace queueTransactions useEffect with a normal function.
2. At a UI level, inform the user about the txs (cannonfile steps) that must be executed outside the safe and show an indicator about the status of each execution.


mjlescano:
closing due to inactivity

PR #1508, opened 2024-10-25
fix(builder): settings must have a defined value

saeta-eth:
This PR aims to fix a bug where settings could be defined without a value. Previously, we didn’t catch this, resulting in default values being used instead of throwing an error.


mjlescano:
@dbeal-eth for this one we needed for you to set the missing settings on snx deployment files, to avoid breakages and to be able to run our tests successfully 🙏

dbeal-eth (changes requested):
can you add a test to confirm the specific case that this fix is applying for?

PR #1509, opened 2024-10-30
fix(website): Optimizations to prevent CPU spikes

saeta-eth:
When the deployment state is too large—like the SNX Arbitrum deployment with its 300k-line JSON file—it slowed down the Queue Deployment page by processing on the main thread. The solution now uses Web Workers, keeping the main thread free and preventing CPU spikes.

Closes https://linear.app/usecannon/issue/CAN-610/performance-issue-on-builder-page-on-snx-arbitrum-build


packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -91,41 +94,68 @@ function useMergedCannonDefInfo(
...
+      } catch (e) {
+        // fallback to non-worker execution if worker fails
+        return new ChainDefinition(deployInfo as RawChainDefinition);
+      }
+      return null;
+    },
+    enabled: Boolean(
+      partialDeployInfo?.ipfsQuery.data?.deployInfo || originalCannonDefInfo.def
+    ),
+  });
```
mjlescano, line 144: This should be moved to a separate hook to avoid making this file even bigger

packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -226,7 +256,7 @@ export default function QueueFromGitOps() {
     const version = 'latest';
     const preset = cannonDefInfo?.def.getPreset(ctx);
     setPreviousPackageInput(`${name}:${version}@${preset}`);
-  }, [cannonDefInfo?.def, selectedDeployType]);
+  }, [cannonDefInfo, cannonDefInfo?.def, selectedDeployType]);
```
mjlescano, line 259: shouldn't `cannonDefInfo` should be converted to a string in some way? to avoid unnecessary re-renders in case the variable reference changes but still has the same value

packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -474,113 +516,247 @@ export default function QueueFromGitOps() {
...
+  }, [
+    isConnected,
+    openConnectModal,
+    chainId,
+    currentSafe?.chainId,
+    switchChainAsync,
+    doBuild,
+    cannonDefInfo?.def,
+    partialDeployInfo?.ipfsQuery.data?.deployInfo,
+    toast,
```
mjlescano, line 529: Regardless the warning from the linter, which is just a reminder, everything should not be put on the hook state change array, e.g `openConnectModal` is not necessary and maybe triggers unnecessary re-renders.

dbeal-eth (approved):
Fantasmic

PR #1510, opened 2024-10-31
fix(website): button cannot be disabled without a reason

dbeal-eth:
so annoying lol


dbeal-eth:
i would really lik eto merge this or something like it today, but seems like its not a priority. lmk if there is anything we can do @noahlitvin

PR #1511, opened 2024-10-31
feat(website): redesign

noahlitvin:


packages/cli/src/chains.ts
```diff
@@ -1,18 +1,7 @@
...
-  nativeCurrency: {
-    name: 'Ether',
-    symbol: 'ETH',
-    decimals: 18,
-  },
-  rpcUrls: { default: { http: ['http://localhost:8545'] } },
-};
-
-export const chains: viem.Chain[] = [cannonChain, ...Object.values(viemChains)];
+export const chains: viem.Chain[] = [...Object.values(viemChains)];
```
mjlescano, line 4: Why this change?

noahlitvin: They added cannon to the viem list so it's redundant now

PR #1521, opened 2024-11-04
fix(website): manually estimate gas from cost of individual txns

dbeal-eth:
rather than trying to use the regular `estimateGas` facilities which can be unreliable, we compute the total gas that should be required by adding the component cost of the multicall txns and expected costs of safe execution and multicall. the sum is the gas estimate, which should be much more accurate

### DIFF
--- a/packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
+++ b/packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
@@ -402,7 +402,7 @@ export default function QueueFromGitOps() {
   let totalGas = BigInt(0);
 
   for (const step of buildState.result?.safeSteps || []) {
-    totalGas += BigInt(step.gas.toString());
+    totalGas += BigInt(step.gas.toString()) + BigInt(50000);
   }
 
   const toast = useToast();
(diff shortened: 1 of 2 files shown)


dbeal-eth:
one caveat is that if the network that we are interacting with has much different EVM usage costs for particular operations (ex. if a SLOAD takes more gas units then it does not ethereum mainnet), the estimate could be off

generally speaking this shouldn't be a big problem on most networks and would be absorbed into the estimate, but we want to be careful.

packages/website/src/features/Deploy/QueueFromGitOpsPage.tsx
```diff
@@ -402,7 +402,7 @@ export default function QueueFromGitOps() {
   let totalGas = BigInt(0);
 
   for (const step of buildState.result?.safeSteps || []) {
-    totalGas += BigInt(step.gas.toString());
+    totalGas += BigInt(step.gas.toString()) + BigInt(50000);
```
mjlescano, line 405:
```suggestion
    totalGas += BigInt(step.gas.toString()) + BigInt(50_000);
```

packages/website/src/hooks/backend.ts
```diff
@@ -240,6 +240,7 @@ export function useTxnStager(
       safeTxn.refundReceiver,
       '0x' + execSig.map((s) => s.slice(2)).join(''),
     ],
+    gas: BigInt(Math.floor((parseInt(safeTxn.safeTxGas) + 500000) * 1.1)),
```
mjlescano, line 243:
```suggestion
    gas: BigInt(Math.floor((parseInt(safeTxn.safeTxGas) + 500_000) * 1.1)),
```

mjlescano (approved):

dbeal-eth:
update: this change does not seem to be applying to the actual txn

PR #1536, opened 2024-11-08
feat(cli): add render foundry script

dbeal-eth:
it writes a script. there are a few benefits with how the script is constructed beyond just the minimal:
* all contracts and function calls are executed with bytecode, meaning that within the scope of a single packge deployment, the script is 100% portable--ie it can be simply copied into a foundry project and used directly as-is.
* the script has a function `getAddress` which can be used by foundry tests to introspect the deployed contracts.

to use, run in a foundry project: `cannon build --wipe --write-script script/Deploy.sol --write-script-format foundry`.

hardhat project works too.

note that the previous `foundry` script generator which creates a shell script for cast has been renamed to `cast`.

when testing against synthetix, there are some minor issues that can come up when using `import` that are probably best dealt with manually in the repo (these issues would not be present for example in the omnibus repo).

when testing against omnibus, it nailed it. generates a deployment script that can be executed with foundry broadcast as far as I could tell (didnt actually run on testnet just yet, just experimentation)

fixes https://linear.app/usecannon/issue/CAN-598/add-write-script-functionality-using-solidity-scripts


packages/cli/src/write-script/render-cast.ts
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * This script is used to deploy contracts using Foundry Cast.
+ * It outputs a bash script that can be used to deploy contracts and execute transactions.
+ * Note: Make sure you add `.sh` extension to the output file.
+ */
+
+const header = `#!/bin/bash
+
+# Set your environment variables
+# Set your environment variables
```
mjlescano, line 13:
```suggestion
```

PR #1572, opened 2024-12-03
fix(website): version dialog preset

AkihisaY:
Added the tooltips next to "Preset" in the table header
Also fixed z tag to prevent tooltip content under a modal

Screenshot 2024-12-03 at 15 26 38

### DIFF
--- a/packages/website/postcss.config.js
+++ b/packages/website/postcss.config.js
@@ -3,4 +3,4 @@ module.exports = {
     tailwindcss: {},
     autoprefixer: {},
   },
-}
+};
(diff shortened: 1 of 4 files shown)


noahlitvin (commented):
Thanks for this!
* What's the scenario where changing the z-index on the tooltip necessary? I'd rather leave this as default if possible.
* Can we put a max width on this tooltip (but not the default)?
* Can we add something like `align-center` or (`justify-center`?) to better align the icon with the other items in the header cell?

AkihisaY:
Thank you for confirming that. I put a max with and item-center.

Why I fixed the z-index on the tooltip is because it showed up under the modal because z-index of a modal was 1400 and tooltip's was 50. Then I thought that it was better to fix the default on the tooltip so that nobody has the same issue ever. I'd be happy if you could give me what is the best and some advice. I could fix that again.

PR #1576, opened 2024-12-05
fix(website): copy button component

AkihisaY:
Added the clipboard component
Had the 0x0000... hash if the item is empty


dbeal-eth (approved):
@noahlitvin could you check this too

PR #1579, opened 2024-12-06
fix(website): update guides with AI assisted improvements

dbeal-eth:
I did a pass afterwards and they seem pretty good changes to me, but please validate me.

### DIFF
--- a/packages/website/guides/build-with-cannon.mdx
+++ b/packages/website/guides/build-with-cannon.mdx
@@ -13,6 +13,8 @@ after:
 
 ## Build with Cannon
 
+In the previous guide, we showed how to setup a basic Solidity project and add Cannon to the project. This guide explains how to build and run your project using the `cannon build` and `cannon run` commands.
+
 <Tabs defaultValue="foundry">
   <TabsList className="w-full">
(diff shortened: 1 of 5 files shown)


packages/website/guides/build-with-cannon.mdx
```diff
@@ -24,195 +26,218 @@ after:
...
+
+   Run this package
+   > cannon learn-cannon:0.0.1
+   ```
+
+   <Alert variant="info">
+     <AlertDescription>
+       If you'd like to keep the node running after the build process, add the `--keep-alive` flag to the build command. For example: `cannon build --keep-alive`.
+     </AlertDescription>
+   </Alert>
```
mjlescano, line 101: IMHO, I would remove this alert here and just leave it in the API docs. As this guide is intended for newish users and the --keep-alive is an super specific case, and users normally don't know whats the "build node" to keep running, etc

```suggestion
```

mjlescano (approved):

packages/website/guides/build-with-cannon.mdx
dbeal-eth: sorry missed this here, can do it later

PR #1580, opened 2024-12-08
feat(ipfs): add indexing of published packages to S3

mjlescano:


packages/indexer/src/pinner.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    const rawData = await readRawIpfs({
+      ipfsUrl: config.IPFS_URL,
+      cid,
+    });
+
+    await s3.putObject(cid, rawData);
+  },
+
+  async PIN_PACKAGE(data: { cid: string }) {
```
dbeal-eth, line 30: this segment of code is not sufficient to ensure that a package is published. You must also pin any subpackages (ie, `pull`ed or `clone`d package urls must also be pinned, and they may not be pinned automatically!)

dbeal-eth (changes requested):

dbeal-eth (approved):

PR #1582, opened 2024-12-10
fix(builder): Disable extraneous check

dbeal-eth:


AkihisaY (approved):
I tested it on the test site.

PR #1589, opened 2024-12-10
fix(builder): improve 2.20 speed significantly

dbeal-eth:
works by removing unnecessary json srtingification and using arrays for stable comparison instead. uses generator function to minimize unnecessary computation.

this is critically important when you have synthetix scale deployments with hundreds of steps.

### DIFF
--- a/packages/builder/src/steps/invoke.ts
+++ b/packages/builder/src/steps/invoke.ts
@@ -330,7 +330,7 @@ const invokeSpec = {
       [
         cfg.target?.map((t) => getContractFromPath(ctx, t)?.address), // to
         cfg.func, // func
-        cfg.args?.map((v) => JSON.stringify(v)), // args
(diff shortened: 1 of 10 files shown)


mjlescano (approved):

packages/builder/src/steps/invoke.ts
```diff
@@ -327,6 +327,14 @@ const invokeSpec = {
     }
 
     return [
+      [
+        cfg.target?.map((t) => getContractFromPath(ctx, t)?.address), // to
+        cfg.func, // func
+        cfg.args?.map((v) => JSON.stringify(v)), // args
```
mjlescano, line 333: Very small speed improv, but
```suggestion
        cfg.args ? JSON.stringify(cfg.args) : null, // args
```

PR #1591, opened 2024-12-11
auto retry failed txn

dbeal-eth:
if a transaction fails to execute, auto retry to ensure that execution continues

later on we would want this to be an actual retry button


mjlescano:
@dbeal-eth should we convert this one in draft?

mjlescano:
Will close this as it is not necessary anymore

PR #1593, opened 2024-12-12
fix(cli): remove @ipfs: protocol compatibility

mjlescano:
The protocol format was intended from the beginning to have custom storage protocols, but we ended up deciding on not doing that and is not necessary anymore.


mjlescano:
This one was already introduced by https://github.com/usecannon/cannon/pull/1594

PR #1594, opened 2024-12-12
fix(cli): allow to use ipfs:// on all commands (that make sense)

mjlescano:
This PR makes the `getPackageInfo` fn to try and parse the given package reference, or, if an ipfsHash/Url is given, it will try to download it and get the fullPackageReference. Also, the chainId it will be taken by the users param `--chain-id`, the package content, the given `--rpc-url` provider, or lastly it will prompt the user to add it.

This functionality was added to the following commands:
* `verify`
* `diff`
* `publish`
* `unpublish`
* `inspect`
* `trace` (now we can trace from ipfs hash also 🙌)
* `decode`
* `interact`

There's a special case for the following commands that a correct `packageReference` is required, and not an ipfs ref. This is now being correctly parsed and if needed the user is prompted for a chainId
* `fetch`

Another special case is on the contrary, when only an `ipfsUrl` is allowed, but now we are improving the validation:
* `pin`

# Extras
1. Added a new `packages/cli/.env.test` file to be able to configure env vars for e2e tests


packages/cli/src/commands/config/index.ts
```diff
@@ -549,7 +547,6 @@ export const commandsConfig: CommandsConfig = {
       {
         flags: '-c --chain-id <chainId>',
         description: 'Chain ID of the variant to inspect',
-        defaultValue: '13370',
```
mjlescano, line 552: On inspect command, is better to ask the user for a chainId

packages/cli/src/util/commands-config.ts
```diff
@@ -16,8 +16,6 @@ export const applyCommandsConfig = (command: Command, config: CommandConfig) =>
     config.arguments.map((argument: any) => {
       if (argument.flags === '<packageRefs...>') {
         command.argument(argument.flags, argument.description, parsePackagesArguments, argument.defaultValue);
-      } else if (command.name() === 'interact' && argument.flags === '<packageRef>') {
-        command.argument(argument.flags, argument.description, parsePackageArguments, argument.defaultValue);
```
mjlescano, line 20: Small refactor, as the `parsePackageArguments` is intended for parsing settings, but in `interact` we don't need to parse settings.

packages/cli/src/commands/fetch.test.ts
mjlescano, line 1: Deleting this test as it is really testing almost nothing. And also we have a proper e2e test that already covers this case.

packages/cli/test/e2e/config/settings.json
```diff
@@ -1,8 +1,6 @@
 {
-  "publishIpfsUrl": "https://us-east.repo.usecannon.com",
+  "publishIpfsUrl": "https://repo.usecannon.com",
   "rpcUrl": "http://127.0.0.1:9545",
   "privateKey": "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
-  "registryRpcUrl": "http://127.0.0.1:9545",
-  "etherscanApiUrl": "https://api.etherscan.io/api",
-  "etherscanApiKey": "A7PNIZCEI2HDRUTDJBWS36XBGIB7Q74YMK"
```
mjlescano, line 7: Removing pasted key 👮‍♂️

packages/cli/src/commands/decode.ts
```diff
@@ -35,14 +33,6 @@ export async function decode({
...
 
   if (isTxHash(data)) {
-    if (!chainId && rpcUrl && isURL(rpcUrl)) {
-      chainId = await getChainIdFromRpcUrl(rpcUrl);
-    } else {
-      if (!chainId) throw new Error('--chain-id or --rpc-url is required to decode transaction data');
-    }
-
-    await ensureChainIdConsistency(rpcUrl, chainId);
-
```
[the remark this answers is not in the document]
mjlescano: Yup, those validations where moved when calling `getPackageInfo` here: https://github.com/usecannon/cannon/blob/fix-ipfs-fetch/packages/cli/src/index.ts/#L616

In turn, `getPackageInfo` calls [`getPackageReference`](https://github.com/usecannon/cannon/blob/fix-ipfs-fetch/packages/cli/src/helpers.ts/#L464-L476) which includes the validations

mjlescano:
Merging as I got in-chat approval both from @saeta-eth and @dbeal-eth

PR #1600, opened 2024-12-19
fix(website): fix well-formatted hex

AkihisaY:
Fixed well-formatted hex on the _packageTags item by following CAN-564.
Also added ByteInput component so that character validation is available like the picture.

Screenshot 2024-12-19 at 11 11 53


packages/website/src/features/Packages/FunctionInput.tsx
```diff
@@ -90,6 +91,14 @@ export const FunctionInput: FC<Props> = ({
       case input.type.startsWith('tuple'):
         // TODO: implement value prop for TupleInput
         return <TupleInput input={input} handleUpdate={_handleUpdate} />;
+      case input.type.startsWith('bytes32'):
```
dbeal-eth, line 94: please add handlers for other bytes types too (bytes1, bytes2, bytes4, bytes8, bytes16)

packages/website/src/features/Packages/FunctionInput/ByteInput.tsx
```diff
@@ -0,0 +1,63 @@
...
+            size="sm"
+            onChange={(e) => {
+              setUpdateValue(e.target.value || '');
+            }}
+            _focus={{
+              borderColor: isInvalid ? 'red.500' : 'blue.300',
+            }}
+          />
+        </TooltipTrigger>
+        <TooltipContent className="max-w-sm text-center">
```
dbeal-eth, line 56: Better UX would probably be to say "inpupt is too short" if its to oshort, or "input is too long" if its too long

packages/website/src/features/Packages/FunctionInput/DefaultInput.tsx
```diff
@@ -10,22 +9,7 @@ export const DefaultInput: FC<{
...
   const [updateValue, setUpdateValue] = useState<string>(value);
   useEffect(() => {
-    const bytes32Regex = /^0x[0-9a-fA-F]{64}$/;
-    if (inputType == 'bytes32' && !bytes32Regex.test(updateValue)) {
-      if (updateValue.startsWith('0x')) {
-        handleUpdate(updateValue);
-      } else {
-        handleUpdate(stringToHex(updateValue, { size: 32 }));
-      }
-    } else if (inputType == 'tuple') {
```
dbeal-eth, line 20: want to make sure tuple inputs work on other parts of the site, will do some testing. otherwise, this is good to see!

dbeal-eth (approved):

packages/website/src/features/Packages/FunctionInput/DefaultInput.tsx
AkihisaY: When I conformed if this affects any other places, this component is called on only one place which is FunctionInput.tsx and it has the condition of "tuple". So "tuple" function on DefaultInput.tsx is never used and we can remove it.

packages/website/src/features/Packages/FunctionInput.tsx
AkihisaY: Added other types.

dbeal-eth (approved):

dbeal-eth:
manual testing looks good :+1:

PR #1601, opened 2024-12-19
fix(builder): template sanitization + sandboxing for untrusted code execution

saeta-eth:


packages/builder/src/utils/template.ts
```diff
@@ -1,8 +1,141 @@
...
+  try {
+    const functionProps = new Function('return Object.getOwnPropertyNames(this)')();
+    functionProps.forEach((prop: string) => globals.add(prop));
+  } catch (e) {
+    // Ignore if blocked
+  }
+
+  // Remove allowed objects like JSON, Number, Math from the global properties
+  // TODO: do we need to whitelist any other objects?
+  const allowedGlobals = ['JSON', 'Number', 'Math', 'String', 'parseFloat', 'parseInt', 'isNaN', 'console'];
```
dbeal-eth, line 59: we will definitely want to check a full list somewhere, but this probably covers the important stuff :+1:

packages/builder/src/utils/template.ts
```diff
@@ -1,8 +1,141 @@
...
+  allowedGlobals.forEach((allowed) => globals.delete(allowed));
+
+  return globals;
+}
 
+/**
+ * List of blocked globals so we can use them in validation
+ */
+const BLOCKED_GLOBALS = new Set([
```
dbeal-eth, line 68: these are keywords not global variables/globals

packages/builder/src/steps/invoke.ts
```diff
@@ -361,66 +361,67 @@ const invokeSpec = {
...
-    config.func = template(config.func)(ctx);
+    config.func = executeTemplate(config.func, ctx, 'ctx');
 
     if (config.args) {
       debug('rendering invoke args with settings: ', ctx.settings);
-      config.args = _.map(config.args, (a) => {
-        // just convert it to a JSON string when. This will allow parsing of complicated nested structures
-        return JSON.parse(template(JSON.stringify(a))(ctx));
+      config.args = _.map(config.args, (arg) => {
+        // TODO: fix this
```
dbeal-eth, line 376: what should be fixed here? not obvious to me perhaps

packages/builder/src/utils/template.ts
```diff
@@ -48,6 +185,39 @@ export function template(str?: string, options?: TemplateOptions) {
...
 }
 
+/**
+ * Executes a template string in a SES secure compartment.
+ * @param templateStr - The template string to evaluate
+ * @param context - The context object containing endowments
+ * @param importKey - The key to use for imports (defaults to 'ctx')
+ * @returns The evaluated result
+ */
+export function executeTemplate(templateStr: string, context: Record<string, any> = {}, importKey: 'ctx' | 'recorders') {
```
dbeal-eth, line 195: we almost always use `ctx` for `importKey`, so we should set the default value to `importKey`

also, if we rename the inner `template` function, we can avoid having to make changes to many files, which simplifies the look of this PR overall

dbeal-eth (changes requested):
great work! I particularly like the effort to add so much documentation throughout many parts of the codebase. Should improve the builder maintainability. amazing.

only a couple minor things to look at before this is mergable.

packages/builder/src/utils/template.ts
mjlescano: done! didn't add absolutely everything, but added some sane defaults using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference as reference

dbeal-eth (approved):

PR #1604, opened 2024-12-27
fix(website): add custom default rpcs

mjlescano:
This PR adds the possibility for changing the default rpcs given by viem. In our case this is necessary because we've been seeing that the default Sepolia rpc does not work properly for the users on the website.


mjlescano:
Because that one is only used for the registry and it is for ETH Mainnet (which is working properly). I modified only the Sepolia default rpc, and because of the way i did it, the user can still use their custom rpc on the settings and it will be overriden.

noahlitvin (approved):

PR #1608, opened 2024-12-30
fix(builder): refactor types for action definitions

mjlescano:


packages/builder/src/actions.ts
```diff
@@ -13,39 +13,39 @@ import diamondSpec from './steps/diamond';
 import varSpec from './steps/var';
 import { ChainArtifacts, ChainBuilderContext, ChainBuilderContextWithHelpers, PackageState } from './types';
 
-export interface CannonAction {
+export interface CannonAction<Config = any> {
   label: string;
 
-  configInject: (ctx: ChainBuilderContextWithHelpers, config: any, packageState: PackageState) => any;
+  configInject: (ctx: ChainBuilderContextWithHelpers, config: Config, packageState: PackageState) => Config;
```
dbeal-eth, line 19: The `config` argument should be a `RawConfig` type or so right?

dbeal-eth (approved):

packages/builder/src/actions.ts
mjlescano: Something like this? https://github.com/usecannon/cannon/pull/1608/commits/cd57a1a4ba64ba41bc1a6a40bc17f454689b4392

I like it :), FYI, by default is an `any` so we don't break the types in case someone is using the type when defining custom actions.

PR #1613, opened 2025-01-02
feat(cli): improve publish command output

noahlitvin:


packages/cli/src/commands/publish.ts
```diff
@@ -138,54 +132,75 @@ export async function publish({
...
+      log(gray(`  Publishing Fee: ${viem.formatEther(totalFees)} ETH`));
+      
+      if (totalFees > 0n) {
+        const balance = await onChainRegistry.provider!.getBalance({ 
+          address: onChainRegistry.signer!.address 
+        });
+        
+        if (totalFees >= balance) {
+          throw new Error('You do not appear to have enough ETH in your wallet to publish');
+        }
```
saturn-dbeal, line 174: The balance check is now inside `if (!quiet)`, meaning in quiet mode the insufficient balance check is skipped. The transaction will still fail, but the user gets a less helpful error. Consider moving the balance check outside the `!quiet` block:

```typescript
if (onChainRegistry instanceof OnChainRegistry) {
  const totalFees = await onChainRegistry.calculatePublishingFee(publishCalls.length);
  
  if (totalFees > 0n) {
    const balance = await onChainRegistry.provider!.getBalance({ 
      address: onChainRegistry.signer!.address 
    });
    
    if (totalFees >= balance) {
      throw new Error('You do not appear to have enough ETH in your wallet to publish');
    }
  }
  
  if (!quiet) {
    log(bold('\nTransaction Details:'));
    log(gray(`  Publishing Fee: ${viem.formatEther(totalFees)} ETH`));
  }
}
```

packages/cli/src/commands/publish.ts
```diff
@@ -138,54 +132,75 @@ export async function publish({
...
-      message: 'Proceed?',
-      initial: true,
+      name: 'proceed',
+      message: 'Proceed with publishing?',
     });
 
-    if (!verification.confirmation) {
-      log('Cancelled');
-      process.exit(1);
+    if (!proceed) {
```
saturn-dbeal, line 186: Good change — `throw new Error('User aborted')` is much better than `process.exit(1)` for testability and proper error handling upstream.

packages/cli/src/commands/publish.ts
```diff
@@ -138,54 +132,75 @@ export async function publish({
...
-    log(blueBright('Transactions:'));
-    for (const tx of registrationReceipts) log(`  - ${tx}`);
+    log(gray('\nPublishing packages...'));
   }
+  
+  const txHashes = await onChainRegistry.publishMany(publishCalls);
+  
+  if (!quiet) {
+    log(`\n${green('✓')} Successfully published packages`);
+    for (const txHash of txHashes) {
```
saturn-dbeal, line 199: Good change — returning the txHashes makes this function more useful for programmatic consumers.

saturn-dbeal (changes requested):
Tests pass (6/6). The output formatting improvements are nice — cleaner structure with separators and better readability. However, there are a few issues that should be addressed:

1. **Balance check moved inside `!quiet` block** — The insufficient balance check is now gated by `if (!quiet)`, meaning in quiet mode the user won't get the early error and will instead get a cryptic transaction failure. The balance check should always run.

2. **Breaking change to function signature** — The `tags` parameter was moved before `onChainRegistry` in the function signature. This is a breaking change for any callers using positional arguments. Not necessarily wrong, but worth documenting if intentional.

3. **Changed from `toStorage.registry.publishMany` to `onChainRegistry.publishMany`** — This bypasses the storage abstraction layer. If `toStorage.registry` was ever a different registry implementation than `onChainRegistry`, this would change behavior. Looks like they're always the same in practice, but the change is worth calling out.

packages/cli/src/commands/publish.ts
saturn-dbeal, line 174: The balance check is now inside `if (!quiet)`, meaning in quiet mode the insufficient balance check is skipped. The transaction will still fail, but the user gets a less helpful error. Consider moving the balance check outside the `!quiet` block:

```typescript
if (onChainRegistry instanceof OnChainRegistry) {
  const totalFees = await onChainRegistry.calculatePublishingFee(publishCalls.length);
  
  if (totalFees > 0n) {
    const balance = await onChainRegistry.provider!.getBalance({ 
      address: onChainRegistry.signer!.address 
    });
    
    if (totalFees >= balance) {
      throw new Error('You do not appear to have enough ETH in your wallet to publish');
    }
  }
  
  if (!quiet) {
    log(bold('\nTransaction Details:'));
    log(gray(`  Publishing Fee: ${viem.formatEther(totalFees)} ETH`));
  }
}
```

packages/cli/src/commands/publish.ts
saturn-dbeal, line 186: Good change — `throw new Error('User aborted')` is much better than `process.exit(1)` for testability and proper error handling upstream.

packages/cli/src/commands/publish.ts
saturn-dbeal, line 199: Good change — returning the txHashes makes this function more useful for programmatic consumers.

saturn-dbeal (changes requested):
Tests pass (6/6). The output formatting improvements are nice — cleaner structure with separators and better readability. However, there are a few issues that should be addressed:

1. **Balance check moved inside `!quiet` block** — The insufficient balance check is now gated by `if (!quiet)`, meaning in quiet mode the user won't get the early error and will instead get a cryptic transaction failure. The balance check should always run.

2. **Breaking change to function signature** — The `tags` parameter was moved before `onChainRegistry` in the function signature. This is a breaking change for any callers using positional arguments. Not necessarily wrong, but worth documenting if intentional.

3. **Changed from `toStorage.registry.publishMany` to `onChainRegistry.publishMany`** — This bypasses the storage abstraction layer. If `toStorage.registry` was ever a different registry implementation than `onChainRegistry`, this would change behavior. Looks like they're always the same in practice, but the change is worth calling out.

saturn-dbeal (changes requested):
Tests pass (6/6). The output formatting improvements are nice — cleaner structure with separators and better readability. However, there are a few issues that should be addressed:

1. **Balance check moved inside `!quiet` block** — The insufficient balance check is now gated by `if (!quiet)`, meaning in quiet mode the user won't get the early error and will instead get a cryptic transaction failure. The balance check should always run.

2. **Breaking change to function signature** — The `tags` parameter was moved before `onChainRegistry` in the function signature. This is a breaking change for any callers using positional arguments. Not necessarily wrong, but worth documenting if intentional.

3. **Changed from `toStorage.registry.publishMany` to `onChainRegistry.publishMany`** — This bypasses the storage abstraction layer. If `toStorage.registry` was ever a different registry implementation than `onChainRegistry`, this would change behavior. Looks like they're always the same in practice, but the change is worth calling out.

saturn-dbeal (changes requested):
Tests pass (6/6). The output formatting improvements are nice — cleaner structure with separators and better readability. However, there are a few issues that should be addressed:

1. **Balance check moved inside `!quiet` block** — The insufficient balance check is now gated by `if (!quiet)`, meaning in quiet mode the user won't get the early error and will instead get a cryptic transaction failure. The balance check should always run.

2. **Breaking change to function signature** — The `tags` parameter was moved before `onChainRegistry` in the function signature. This is a breaking change for any callers using positional arguments. Not necessarily wrong, but worth documenting if intentional.

3. **Changed from `toStorage.registry.publishMany` to `onChainRegistry.publishMany`** — This bypasses the storage abstraction layer. If `toStorage.registry` was ever a different registry implementation than `onChainRegistry`, this would change behavior. Looks like they're always the same in practice, but the change is worth calling out.

PR #1621, opened 2025-01-07
fix(cli): retries to IPFS were not being done

mjlescano:
Noticed that the default configuration for the `axios-retry` library we were using for making retries when a IPFS POST request fail was not triggered on `POST` requests, this PR fixes that.

And, as an extra retry measure, retries were added when making S3 requests on repo.

### DIFF
--- a/packages/builder/src/ipfs.ts
+++ b/packages/builder/src/ipfs.ts
@@ -82,10 +82,15 @@ export async function prepareRawFormData(data: Buffer) {
   return { cid, formData };
 }
 
+let _currentRetries = 0;
 export function setAxiosRetries(totalRetries = 3) {
+  if (_currentRetries === totalRetries) return;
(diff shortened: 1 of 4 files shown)


packages/builder/src/ipfs.ts
```diff
@@ -82,10 +82,15 @@ export async function prepareRawFormData(data: Buffer) {
...
 
+let _currentRetries = 0;
 export function setAxiosRetries(totalRetries = 3) {
+  if (_currentRetries === totalRetries) return;
+  _currentRetries = totalRetries;
+
   axiosRetry(axios, {
     retries: totalRetries,
     shouldResetTimeout: true,
+    retryCondition: () => true,
```
mjlescano, line 93: This is the most important config change needed.

noahlitvin (approved):

PR #1636, opened 2025-01-17
feat(registry): Add registry subscription model

mjlescano:
Subscription model to the registry with possibility of paying with any ERC20 token implementation.

## FAQs
* `CannonSubscription` is a separate contract, though that it would be easier for future upgrades.
* Subscription Plans are defined by us, and there is going to be always a "default" one for new users.
* Subscription Plans are a combination of Price by Term, and the user when subscribing has to pay `price * amountOfTerms` required.
* We can set the minimum or maximum amount of terms a user can have of a given plan. This means that:
  1. e.g. using the `minTerms` of a plan, we can set that for a plan that has 100 credits per month, the user needs to pay for at least 12 terms, which would mean a 1 year subscription.
  2. e.g. for max duration, the user cannot subscribe more than 2 years in advance (namely 24 terms), this is so we don't commit indefinitely to offering that plan.
* A `Membership` is an instance of a given `Plan` for a `User`, which saves since when it started and its expiration date. It also has tally of `credits` by term.
* The user can call `cancelMembership`, it will delete the existing membership and give back the USDC of the terms that **didn't start yet**. This means if the terms are 1 month, and the user is on month 3 of a 1 year subscription, we will give back 9 months worth of USDC that they payed. (for this to work the registry has to have sufficient USDC, ofc)
* If we create a new default plan, the users that are already subscribed to an older plan they will still have the credits from the old plan.
* We can disable an old plan so the users that are subscribed to them cannot renew them (but they will be able to finish the membership until it ends)

## TODO
- [x] Subscription Plan
- [x] Membership Creation/Update
- [x] Membership Payment
- [x] New Publish logic (count publishes, update membership if necessary)
- [x] Update tests
- [ ] Update cannonfiles
- [ ] Builder/CLI subscription logic
- [ ] Prevent re-publish


packages/registry/contracts/CannonRegistry.sol
```diff
@@ -179,6 +186,29 @@ contract CannonRegistry is EfficientStorage, OwnedUpgradableUpdated {
       revert FeeRequired(store.publishFee);
     }
 
+    _publish(_packageName, _variant, _packageTags, _packageDeployUrl, _packageMetaUrl);
+  }
+
+  function publishWithSubscription(
```
dbeal-eth, line 192: ok so looks like we will need to make sure that we call the right function from the cli?  was hoping it would just automatically detect by sender address or whatever, but I guess this would be fine. either way, an important thing to note.

packages/registry/contracts/CannonSubscription.sol
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * @title Management of Subscriptions to Cannon Registry
+ */
+contract CannonSubscription is ReentrancyGuard {
+  using Subscription for Subscription.Data;
+
+  error ZeroAddressNotAllowed(string variableName);
+  error InsufficientAllowance(address user, uint256 allowance, uint256 required);
+  error TransferFailed();
```
dbeal-eth, line 18: could use args here

packages/registry/contracts/CannonSubscription.sol
```diff
@@ -0,0 +1,157 @@
...
+    Subscription.Data storage _subscription = Subscription.load();
+
+    Subscription.Membership storage _membership = _subscription.getMembership(_sender);
+    Subscription.Plan storage _plan = _membership.planId == 0
+      ? _subscription.getDefaultPlan()
+      : _subscription.getPlan(_membership.planId);
+
+    uint256 _totalPrice = _plan.price * _amountOfTerms;
+
+    if (_totalPrice / _amountOfTerms != _plan.price) {
```
dbeal-eth, line 95: i feel like this could be manipulated somehow with certain values of `_plan.price` to get it to wrap in on itself and get an incredibly large number of terms

I feel like we should cap `_amountOfTerms` to some reasonable number instead (ex. up to 24 terms, 2 years probably)

packages/registry/contracts/CannonSubscription.sol
```diff
@@ -0,0 +1,157 @@
...
+    uint256 _allowance = USDC.allowance(_sender, VAULT);
+    if (_allowance < _totalPrice) {
+      revert InsufficientAllowance(_sender, _allowance, _totalPrice);
+    }
+
+    _subscription.acquireMembership(_plan, _membership, _amountOfTerms);
+
+    bool _success = USDC.transferFrom(_sender, VAULT, _totalPrice);
+    if (!_success) {
+      revert TransferFailed();
```
dbeal-eth, line 108: include here the parameters for the transfer call that failed

dbeal-eth (changes requested):
looks good overall!

I feel like we want to add a couple capabilities:
* user should be able to get a discount for purchasing more than 1 (ex. 12 terms) at the same time. any terms that are discounted are not refundable via `cancelSubscription`
* we should be able to offer discounts for certain users we have other agreements with. admin calls a function `offerDiscount` or something and when the same address calls they can receive the discounted amount

packages/registry/contracts/CannonSubscription.sol
mjlescano: The plans already have a `maxTerms` value, so we are doing this check on subscription.

packages/registry/contracts/CannonSubscription.sol
mjlescano: done!

dbeal-eth (approved):

PR #1645, opened 2025-01-20
fix(website): nested tuples issues

AkihisaY:
Fixed [CAN-681](https://linear.app/usecannon/issue/CAN-681/nested-tuples-still-have-trouble-in-interact-tool-output#comment-faec4b64).
And all outputs are displayed.

Screenshot 2025-01-20 at 15 00 11

### DIFF
--- a/packages/website/src/features/Packages/FunctionOutput.tsx
+++ b/packages/website/src/features/Packages/FunctionOutput.tsx
@@ -34,25 +34,20 @@ const resultText = (
   type: string,
   value: any
 ): string => {
-  // if (name === 'pool' || name === 'startTime' || name === 'epochId') {
-  //   console.log(`startTime or epochId or pool  --> ${name}`);
-  //   console.log(value);
-  // }
   if (value !== null && value !== undefined) {
-    let ret = '';
+    let result = '';
     for (const key in value) {
       for (const idx in value[key]) {
         if (name === idx) {
-          console.log();
-          ret = String(value[key][idx]);
+          result = String(value[key][idx]);
           break;
         }
       }
     }
-    if (type === 'uint256[]' && ret === '') {
+    if (type.includes('[]') && result === '') {
       return '[]';
     }
-    return ret;
+    return result;
   } else if (type === 'string') {
     return '(empty string)';
   } else if (type === 'boolean' || type === 'bool') {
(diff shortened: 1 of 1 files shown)


packages/website/src/features/Packages/FunctionOutput.tsx
```diff
@@ -28,29 +28,99 @@ const ItemLabel: FC<{ name: string; type: string }> = ({ name, type }) => (
...
+    return value;
+  } else if (value !== null && value !== undefined) {
+    const resultItem = value.find(
+      (item: any) => name !== undefined && item.hasOwnProperty(name)
+    );
+    const result: string =
+      resultItem && name !== undefined ? String(resultItem[name]) : '';
+
+    if (type.includes('[]') && result === '') {
+      return '[]';
```
dbeal-eth, line 46: can you describe what this case covers? tbh its a little unclear to me what this does

packages/website/src/features/Packages/FunctionOutput.tsx
```diff
@@ -110,113 +174,69 @@ export const FunctionOutput: FC<{
         const outputValue = value[abiParameter.name];
         return (
           <span className="block pt-1 pb-2 text-xs">
-            {resultText(abiParameter.type, outputValue)}
-            <ClipboardButton
-              text={resultText(abiParameter.type, outputValue)}
-            />
+            {String(outputValue)}
+            <ClipboardButton text={outputValue} />
```
dbeal-eth, line 178: i notice here you changed the clipboard to use the next of `outputValue` rather than `resultText`. Was it not working as expected the other way? seems reasonable but just want to confirm what happened.

dbeal-eth (approved):
also CI checks are failing. please take a lookg :pray:

packages/website/src/features/Packages/FunctionOutput.tsx
AkihisaY: I remember that I added [] before when the output doesn't have any values and the type is array. But if it's not necessary, I'll remove it.

packages/website/src/features/Packages/FunctionOutput.tsx
AkihisaY: In this case, I found out that the value which type is array converts to string and display it like the picture. So I changed it so that it remains.

Screenshot 2025-01-20 at 16 20 28

PR #1648, opened 2025-01-21
fix(website): rawcannofile height issue

AkihisaY:
Fixed the height on Raw Cannonfile tab properly
Also adjusted the codes's width to avoid overlapping with scroll bars.

Screenshot 2025-01-21 at 12 21 47


noahlitvin (changes requested):
I think this is too tall? It should go down to the footer and allow scrolling within it.

mjlescano (approved):

PR #1650, opened 2025-01-22
fix(cli): decode short selector

AkihisaY:
Fixed [CAN-646](https://linear.app/usecannon/issue/CAN-646/decode-should-be-able-to-decode-just-the-selector-without-args)

Screenshot 2025-01-22 at 17 05 07


dbeal-eth (approved):
overall looks good :+1: 

before merging, please double check if the code on lines 208-220 is still needed? seems like its been replaced by the new code on 200-207, but I jus twant to double check.

AkihisaY:
Thanks for confirming that. I confirmed that this code on 200-207 is still used, so I want to leave it.

packages/cli/src/commands/decode.ts
```diff
@@ -193,6 +197,14 @@ function _parseData(abis: ContractData['abi'][], data: viem.Hash) {
   if (!data) return null;
 
   for (const abi of abis) {
+    for (const abiItem of abi) {
+      if (abiItem.type === 'error') {
```
dbeal-eth, line 201: shouldn't we also have `abiItem.type === 'function` here as well?

mjlescano (approved):

PR #1652, opened 2025-01-22
refactor(website): CAN-687 - builder refactor

nicosampler:


mjlescano (changes requested):
When using a git cannonfile url and it doesn't have `deployers =` setted, the "preview txs" button is not appearing

mjlescano (approved):

PR #1657, opened 2025-01-28
fix(cli): process exit code on partial deployments

mjlescano:
Now, when building and finishing with a partial deployment the process exit code is going to be different to `0`.

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -223,7 +223,11 @@ applyCommandsConfig(program.command('build'), commandsConfig.build)
     const pickedCliSettings = _.pick(cliSettings, Object.keys(options));
     const mergedOptions = _.assign({}, options, pickedCliSettings);
 
-    const [node, pkgSpec, outputs, runtime] = await doBuild(cannonfile, settings, mergedOptions);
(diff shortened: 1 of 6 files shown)


packages/cli/src/index.ts
```diff
@@ -223,7 +223,11 @@ applyCommandsConfig(program.command('build'), commandsConfig.build)
     const pickedCliSettings = _.pick(cliSettings, Object.keys(options));
     const mergedOptions = _.assign({}, options, pickedCliSettings);
 
-    const [node, pkgSpec, outputs, runtime] = await doBuild(cannonfile, settings, mergedOptions);
+    const [node, pkgSpec, outputs, runtime, deployInfo] = await doBuild(cannonfile, settings, mergedOptions);
+
+    if (deployInfo.status !== 'complete') {
+      process.exitCode = deployInfo.status === 'partial' ? 10 : 11;
```
noahlitvin, line 229: Do we want 10 and 11 here? Google says 11 is `segmentation fault error` ?

noahlitvin (approved):

packages/cli/src/index.ts
mjlescano: You are right, is more common to use higher numbers for custom errors, updating to 90 🙏

PR #1691, opened 2025-02-12
fix(builder): validation of template accesses which depend on `chainId`

dbeal-eth:
or other globals

this basically turned into a refactor because we have to kind of rethink how we are supplying values to the ChainDefinition. But I think its a move in the right direction.

Unfortunately refactors like this usually lead to breakage, so probably best to save this for next release.

### DIFF
--- a/examples/sample-foundry-project/cannonfile.toml
+++ b/examples/sample-foundry-project/cannonfile.toml
@@ -1,4 +1,4 @@
-name = "greeter-foundry"
+name = "greeter"
 version = "<%= package.version %>"
 description = "Simple project to verify the functionality of cannon"
 keywords = ["sample", "greeter"]
--- a/examples/sample-hardhat-project/cannonfile.consumer.toml
+++ b/examples/sample-hardhat-project/cannonfile.consumer.toml
@@ -15,7 +15,7 @@ options.salt = "second"
 options.msg = "a message from second greeter set"
 
 [invoke.do_change_greeting2]
-target = ["greeters.Greeter"]
+target = ["greeters.Greeterz"]
 func = "setGreeting"
 args = ["<%= settings.change_greeting2 %>"]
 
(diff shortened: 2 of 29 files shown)


examples/sample-hardhat-project/cannonfile.consumer.toml
```diff
@@ -15,7 +15,7 @@ options.salt = "second"
 options.msg = "a message from second greeter set"
 
 [invoke.do_change_greeting2]
-target = ["greeters.Greeter"]
+target = ["greeters.Greeterz"]
```
mjlescano, line 18: Is this ok?

packages/builder/src/steps/pull.ts
```diff
@@ -92,21 +92,29 @@ const pullSpec = {
...
         [importLabel]: {
           url: (await runtime.registry.getUrl(source, chainId))!, // todo: duplication
-          ...(await getOutputs(runtime, new ChainDefinition(deployInfo.def), deployInfo.state))!,
+          ...(await getOutputs(
+            runtime,
+            new ChainDefinition(deployInfo.def, false, {
+              chainId,
+              timestamp: deployInfo.timestamp || 0,
+              package: { version: '0.0.0' },
+            }),
```
mjlescano, line 115: Having to create this dummy object on several places just to be able to instantiate looks prone to error, maybe a helper function to create the ChainDefinition would be useful?

examples/sample-foundry-project/cannonfile.toml
```diff
@@ -1,4 +1,4 @@
-name = "greeter-foundry"
+name = "greeter"
 version = "<%= package.version %>"
 description = "Simple project to verify the functionality of cannon"
 keywords = ["sample", "greeter"]
```
mjlescano, line 4: Would be good to add an action on this or other of the cannonfiles used on the test with Inifinex's case, was is with something like `<%= settings.smth['VALUE_' + chainId]`?

mjlescano:
Something appears to be fishy? socket reports a lot of dependencies changes here: https://github.com/usecannon/cannon/pull/1691#issuecomment-2652574948, but there are no changes on any `package.json`

packages/builder/src/steps/pull.ts
mjlescano: Oh, I see that those are overrides, but are they necessary? looks like they are only used for giving them to the access recorder, in that case wouldn't be enough to just send the dummy object? Why is there an option to send custom overrides?

examples/sample-hardhat-project/cannonfile.consumer.toml
dbeal-eth: i was testing something; probably should be reverted back though

mjlescano:
Closing as appears to be stale

PR #1705, opened 2025-02-20
hotfix(builder): add preset length validation

mjlescano:
Preset length was not being validated, this causes that when generating the variant for publishing to the registry it can end up being more than 32bytes. Luckily, this fix is not going to break any already published package as the current preset max length published is `24bytes`.

This is the original error that was being thrown when trying to pin/publish:
```
cannon pin ipfs://Qm...
Uploading package data for pinning...
Error: SizeOverflowError: Size cannot exceed 32 bytes. Given size: 36 bytes.

Version: viem@2.23.2
    at FallbackRegistry.getUrl (/Users/**/.nvm/versions/node/v20.10.0/lib/node_modules/@usecannon/cli/node_modules/@usecannon/builder/dist/src/registry.js:147:23)
    at async pinIpfs 
```

And now, with the fix it doesn't even allow you to build the package:
Screenshot 2025-02-19 at 22 50 51

### DIFF
--- a/packages/builder/jest.config.ts
+++ b/packages/builder/jest.config.ts
@@ -93,7 +93,7 @@ export default {
   // moduleNameMapper: {},
 
   // An array of regexp pattern strings, matched against all module paths before considered 'visible' to the module loader
-  // modulePathIgnorePatterns: [],
+  modulePathIgnorePatterns: ['/dist/'],
 
   // Activates notifications for test results
   // notify: false,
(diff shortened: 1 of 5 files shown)


mjlescano:
An alpha release was published on [v2.21.5-alpha.0](https://www.npmjs.com/package/@usecannon/cli/v/2.21.5-alpha.0)

noahlitvin (approved):

PR #1706, opened 2025-02-20
fix(website): cypress code coverage

AkihisaY:
- Set up the environment for code coverage
- Fixed test cases in "Interact.feature" and "StageTxns.feature"
- Added "data-testid"
- Replaced some codes with the ClipboardButton component


packages/website/cypress/step-definitions/base.ts
```diff
@@ -5,23 +5,19 @@ Given('User opens the {string} page', (path: string) => {
...
+  cy.contains('a', `${link}`).click();
 });
 
-When('User clicks on the button with {string} {string}', (property: string, label: string) => {
-  cy.get(`button[${property}="${label}"]`).click();
-});
-
-When('User clicks on the {string} element with text {string}', (element: string, text: string) => {
-  cy.get(element).contains(text).click();
+When('User clicks on the {int}st/nd/rd/th button or link with id {string}', (idx: number, id: string) => {
```
mjlescano, line 19: I would change "button or link" for just "element", IMHO it generates more readable tests
```suggestion
When('User clicks on the {int}st/nd/rd/th element with id {string}', (idx: number, id: string) => {
```

packages/website/cypress/step-definitions/base.ts
```diff
@@ -32,6 +28,14 @@ When('User types {string} in the input with placeholder {string}', (text: string
   cy.get(`input[placeholder="${placeholder}"]`).type(text);
 });
 
+When('User types {string} into the {int}st/nd/rd/th input with id {string}', (text: string, idx:number, id: string) =>{
+  cy.get(`[data-testid="${id}"]`).eq(idx - 1).clear().type(`${text}`);
```
mjlescano, line 32: Is it possible for this step to add the `input` type? This is just to be consistent with the text that says `"... into the input ..."`
```suggestion
  cy.get(`input[data-testid="${id}"]`).eq(idx - 1).clear().type(`${text}`);
```

mjlescano (approved):
Great work @AkihisaY! Looks like there's an error on the pnpm-lock file, I think it can be fixed just by running `pnpm i`. Approved aside from that.

packages/website/cypress/step-definitions/base.ts
AkihisaY: Fixed it as your suggestion!

Screenshot 2025-02-24 at 10 47 58

packages/website/cypress/step-definitions/base.ts
AkihisaY: Added `input` as your suggestion.

Screenshot 2025-02-24 at 10 51 37

PR #1715, opened 2025-02-26
fix: 404 link in `packages/cli/src/util/wait-for-event.ts`

rebustron:
Hi! I fixes a 404 link in the wait-for-event.ts file. The original link pointing to Optimism's documentation for L1-to-L2 messaging was outdated and has been updated to the correct URL.


dbeal-eth:
it looks like the link is no longer 404. as a result, going to close

PR #1724, opened 2025-03-02
feat: codecov

dbeal-eth:


packages/website/package.json
```diff
@@ -11,7 +11,8 @@
     "start": "next start",
     "lint": "next lint --fix",
     "serve": "serve out -c ../serve.json -p 3000 > /dev/null 2>&1 &",
-    "test": "vitest",
+    "test:fast": "vitest",
+    "test": "pnpm e2e:headless",
```
mjlescano, line 15: This change will remove the run of unit tests on CI, shouldn't we do it like this to run the e2e on the e2e task?:
```suggestion
    "test": "vitest",
    "test-e2e": "pnpm e2e:headless",
```

PR #1748, opened 2025-03-30
chore(website): delete .zip

nicosampler:


dbeal-eth (changes requested):
technically the IPFS hash url replacement strategy is unnecessary if accessing the cannon website by bafy hash domain (bafy hash is just a different conversion of the Qm hash with lowercase only for domain access)

mjlescano (approved):

PR #1754, opened 2025-04-09
fix(website): Interact not loading

nicosampler:
Fixes the issue with the reverted commit.


packages/website/src/pages/packages/[name]/[tag]/[variant]/interact/index.tsx
```diff
@@ -4,7 +4,7 @@ import { NextSeo } from 'next-seo';
 import defaultSEO from '@/constants/defaultSeo';
 import { usePackageNameTagVersionUrlParams } from '@/hooks/routing/usePackageVersionUrlParams';
 import { useCannonChains } from '@/providers/CannonProvidersProvider';
-import InteractTab from '@/features/Packages/Tabs/InteractTab';
+import InteractComponent from '@/features/Packages/Interact';
```
Comment (nicosampler) [inline packages/website/src/pages/packages/[name]/[tag]/[variant]/interact/index.tsx:7]: This was the issue on the commit we had to revert. The `Interact` component was not being imported and there was an infinite recursive call to the existing function already defined in this file also called `Interact`.

PR #1756, opened 2025-04-10
fix(website): fixed display on deployment

AkihisaY:
Fixed [CAN-148](https://linear.app/usecannon/issue/CAN-148/show-which-contracts-were-deployed-with-create2) and [CAN-734](https://linear.app/usecannon/issue/CAN-734/better-contract-display)
Created the file on lib folder to avoid duplicate functions and moved some functions in it from "interact.tsx".

Screenshot 2025-04-10 at 13 21 29


packages/website/src/features/Packages/DeploymentExplorer.tsx
```diff
@@ -14,18 +13,125 @@ import ContractsTab from './Tabs/ContractsTab';
...
+          path: `/packages/${name}/${tag}/${variant}/interact/${item.moduleName}/${item.contractName}/${item.contractAddress}`,
+          deployType: getDeployType(
+            processedDeploymentInfo,
+            contractState[address]?.deployedOn.toString() || item.moduleName
+          ),
+        };
+      })
+      .filter((item) => item !== undefined);
+
+    setContractStateData(
```
dbeal-eth, line 116: doing this type of `setContractStateData` in a `useEffect` could be dangerous and lead to render loops. It would be better to wrap all this into a `useMemo` instead, which will efficiently repopulate the data whenever it changes, but not trigger render state changes.

dbeal-eth (changes requested):
beautiful job! lets just check the `useEffect` comment that I sent.

packages/website/src/features/Packages/DeploymentExplorer.tsx
```diff
@@ -14,18 +13,125 @@ import ContractsTab from './Tabs/ContractsTab';
...
+      contractAllData.sort((a, b) => {
+        if (a.moduleName === name && b.moduleName !== name) return -1;
+        if (a.moduleName !== name && b.moduleName === name) return 1;
+
+        const valueA: string = a['contractName'];
+        const valueB: string = b['contractName'];
+        return valueA.localeCompare(valueB);
+      })
```
nicosampler, line 124: you could move this logic to a separate function, to make the code more readable on the useEffect.

packages/website/src/features/Packages/DeploymentExplorer.tsx
```diff
@@ -14,18 +13,125 @@ import ContractsTab from './Tabs/ContractsTab';
...
+  return config;
+}
 
 export const DeploymentExplorer: FC<{
   pkg: ApiPackage;
-}> = ({ pkg }) => {
+  name: string;
+  tag: string;
+  variant: string;
+}> = ({ pkg, name, tag, variant }) => {
```
nicosampler, line 74: there are other pages where we consume all this information (coming from the URL) defining safe hooks. Perhaps you can go one step further and do the same. 
You can take a look on how they were implemented here: `packages/website/src/hooks/routing`

packages/website/src/features/Packages/Interact.tsx
```diff
@@ -344,11 +222,11 @@ const Interact: FC = () => {
                   const option = [...highlightedOptions, ...otherOptions].find(
                     (opt) =>
                       opt.moduleName === moduleName &&
-                      opt.contractName === contractName,
+                      opt.contractName === contractName
                   );
                   if (option) {
                     void router.push(
-                      `/packages/${name}/${tag}/${variant}/interact/${option.moduleName}/${option.contractName}/${option.contractAddress}`,
+                      `/packages/${name}/${tag}/${variant}/interact/${option.moduleName}/${option.contractName}/${option.contractAddress}`
```
nicosampler, line 229: we are repeating this URL formation many times, could we have a reusable helper for it?

packages/website/src/features/Packages/DeploymentExplorer.tsx
AkihisaY: Thanks for the review! I created "sortByModulePriority" function.

packages/website/src/features/Packages/DeploymentExplorer.tsx
AkihisaY: As your suggestion, I removed these parameters and call the function to get them.

packages/website/src/features/Packages/Interact.tsx
AkihisaY: I created "buildInteractPath" function to reuse it.

packages/website/src/features/Packages/DeploymentExplorer.tsx
AkihisaY: Thanks for the review. I wrapped all this into useMemo and made sure it worked well.

dbeal-eth (approved):
possibly an opporitunity to continue cleaning up the `Interact.tsx` page, but overall looking good

nicosampler (approved):

dbeal-eth:
going to admin merge because the failing test is known to be working fine on local, but I cannot for the life of me understand why its failing in CI despite all efforts.

PR #1758, opened 2025-04-18
refactor(website): Interact-refactor-2

nicosampler:
Abi component.


packages/website/src/features/Packages/interact/ContractsList.tsx
```diff
@@ -109,19 +139,55 @@ export const ContractsList: FC<ContractsListProps> = ({
...
                       );
                     }}
-                    data-testid={`${option.moduleName}.${option.contractName}-button`}
+                    data-testid={`${option.contractName}-button`}
                   >
-                    <span className="text-sm">
-                      {`${option.moduleName}.${option.contractName}`}
+                    <span className="text-sm font-mono">
+                      <span>{option.contractName}</span>
+                      <span className="text-xs text-muted-foreground">{`.${option.moduleName}`}</span>
```
mjlescano, line 156: Shouldn't the `contractName` be after the module? It should be read for example `synthetix.CoreProxy` instead of `CoreProxy.synthetix`
```suggestion
                      <span className="text-xs text-muted-foreground">{`${option.moduleName}.`}</span>
                      <span>{option.contractName}</span>
```

mjlescano (approved):

packages/website/src/features/Packages/interact/ContractsList.tsx
nicosampler: changed

PR #1759, opened 2025-04-20
remove indefinite CI test that nobody can figure out

dbeal-eth:
many have tried to fix this failing CI test over the last few days with no success. At some point we have to admit defeat and not let this derail the entire operation.


mjlescano (approved):
LGTM

PR #1762, opened 2025-04-21
fix(website): fix interact serachbar

AkihisaY:
Fixed [CAN-729](https://linear.app/usecannon/issue/CAN-729/search-filter-for-contracts-on-website-is-flawed).
When hitting the enter key in the searchbar, the first contract of lists is selected.

Screenshot 2025-04-21 at 11 26 32

### DIFF
--- a/packages/website/src/features/Packages/interact/ContractsList.tsx
+++ b/packages/website/src/features/Packages/interact/ContractsList.tsx
@@ -1,4 +1,4 @@
-import { FC, useState } from 'react';
+import { FC, useState, useEffect } from 'react';
 import { useRouter } from 'next/router';
 import { Tabs, TabsList, TabsTrigger } from '@/components/ui/tabs';
 import {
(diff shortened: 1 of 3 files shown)


dbeal-eth:
great work! this is already a huge step better in user experience!

this is a bit ancillary, but here I can see that the contracts are sorted alphabetically by contract name. This is good! but its a bit confusing just loooking at it from away 
image

could we change this so that the contract name comes first, and then in perhaps smaller, gray text, the subpackage its part of is displayed?

not sure if we should do this same thing on the main bar or not.

other than that, is there any way we can make it so that there is some animation as the list of contracts in the horizontal bar changes? it may be hard to see/notice a change when typing in text if you are not paying close attention.

finally, is it too slow for the browser if we literally re-run the filter every single time a user types a new letter? having a delay when typing kind of makes teh experience feel slow (even though it is not!)

packages/website/src/features/Packages/interact/ContractsList.tsx
```diff
@@ -91,42 +117,65 @@ export const ContractsList: FC<ContractsListProps> = ({
...
+                          option.moduleName,
+                          option.contractName,
+                          option.contractAddress
+                        )
+                      );
+                    }}
+                    data-testid={`${option.contractName}-button`}
+                  >
+                    <span className="text-sm">
+                      {`${option.moduleName}.${option.contractName}`}
```
dbeal-eth, line 145: so yea here lets display `option.contractName` first, and hten in separate, smaller, gray text, the module name

nicosampler (approved):
nice job!

PR #1774, opened 2025-04-28
fix(website): reset default per section

AkihisaY:
Fixed [CAN-713](https://linear.app/usecannon/issue/CAN-713/setting-page-reset-default-per-section)
Created reset setting button component

Screenshot 2025-04-28 at 10 23 38


dbeal-eth:
next we probably need to add a button for `Export settings` and `Import settings` as its often the case we want to copy settings from one instance to another, but that can wait.

dbeal-eth (approved):

PR #1777, opened 2025-05-14
prevent browser from pushing path to history when

dbeal-eth:
use `replace` instaed.

not using this makes it difficult/impossible to use the back button to go back.

https://linear.app/usecannon/issue/CAN-679/infinex-requests


AkihisaY (approved):
I rather leaned it. Thanks!

Issue #1778, opened 2025-05-15
Cannon build fails when event parameter is unnamed


guidanoli:
## Steps

1. Initialize a new Forge project
2. Create a `src/CounterFactory.sol` file with the following contents:

```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

import {Counter} from "./Counter.sol";

contract CounterFactory {
    event CounterCreated(Counter);

    function newCounter() external {
        emit CounterCreated(new Counter());
    }
}
```

3. Create a `cannonfile.toml` file with the following contents:

```toml
name = 'counter-factory'
version = '0.0.1'

[deploy.CounterFactory]
artifact = 'CounterFactory'

[invoke.newCounter]
target = 'CounterFactory'
func = 'newCounter'
args = []
factory.Counter.artifact = 'Counter'
factory.Counter.event = 'CounterCreated'
factory.Counter.arg = 0
```

4. Run `cannon build`.

## Expected result

The package would build normally, just as if the parameter were named.

## Obtained result

```
Executing [invoke.newCounter]...
Error: address is not valid in Counter. Ensure "arg" parameter is correct
An error occured during build. A file with comprehensive information pertaining to this error has been written to /home/guilherme/.local/share/cannon/dumps/2025-05-15T22:01:48.452Z.json. Please include this file when reporting an issue.
    at importTxnData (/home/guilherme/.local/share/pnpm/global/5/.pnpm/@usecannon+builder@2.21.5_typescript@5.4.3/node_modules/@usecannon/builder/dist/src/steps/invoke.js:165:23)
    at Object.exec (/home/guilherme/.local/share/pnpm/global/5/.pnpm/@usecannon+builder@2.21.5_typescript@5.4.3/node_modules/@usecannon/builder/dist/src/steps/invoke.js:453:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /home/guilherme/.local/share/pnpm/global/5/.pnpm/viem@2.28.0_typescript@5.4.3_zod@3.24.3/node_modules/viem/_cjs/utils/promise/withTimeout.js:21:25
    at build (/home/guilherme/.local/share/pnpm/global/5/.pnpm/@usecannon+cli@2.21.5_typescript@5.4.3/node_modules/@usecannon/cli/dist/src/commands/build.js:265:26)
[...]

PR #1779, opened 2025-05-17
fix(builder): event information not parsed properly for factory

dbeal-eth:
its a bit wierd how viem handles it, converting to an array if *any* of the fields are unnamed.

fixes https://github.com/usecannon/cannon/issues/1778


AkihisaY (approved):
All good! Seems to work well.

PR #1790, opened 2025-05-30
feat(indexer): integrate 4byte directory

dbeal-eth:
add an indexer task to pull in the 4 byte directory every now and hten.

once the 4byte directory is pulled in, we are able to decode just about anything in the ethereum space.


dbeal-eth:
will force merge because @AkihisaY has sort of reviewed this as part of his branch that he has been working on

PR #1797, opened 2025-06-05
hotfix(website): contractTransaction

nicosampler:
Also send value for walletClient.sendTransaction


dbeal-eth:
have we been able to verify with the client that th efix works as intended? also e2e tests are failing.

PR #1807, opened 2025-06-24
feat(website): New Address Page

AkihisaY:
Create the new address page based on the [CAN-743](https://linear.app/usecannon/issue/CAN-743/new-address-page-which-can-show-any-address-in-block-explorer-fashion)


dbeal-eth (approved):
been working closely with aki on this so even though I havent looked through every line carefully this is definitely ready for merge

PR #1821, opened 2025-07-27
feat(builder): chain specific build steps

dbeal-eth:
prevent steps from executing unless the specified network is currently being built

this is often useful if you are working with bridge contracts or anything multi-network, as it allows for steps to be quickly excluded without having to create any new files.


dbeal-eth:
@claude please review

PR #1826, opened 2025-08-13
feat(website): add pagination feature

AkihisaY:
add transaction list page and pagination feature based on [CAN-760](https://linear.app/usecannon/issue/CAN-760/address-implement-pagination-for-transactions)

### DIFF
--- a/packages/website/src/hooks/useBlockPages.ts
+++ b/packages/website/src/hooks/useBlockPages.ts
@@ -71,8 +71,7 @@ export function useBlockPages(apiUrl: string, address: string) {
           totalPages: isLastPage ? newPages.length + 1 : null,
           totalTxs,
         });
-      } catch (err) {
-        console.error('Failed to fetch page boundaries:', err);
+      } catch {
         setIsError(true);
       } finally {
         setIsLoading(false);


packages/website/src/hooks/useBlockPages.ts
```diff
@@ -0,0 +1,46 @@
+import { useQuery } from '@tanstack/react-query';
+import { searchTransactions } from '@/lib/address';
+import { MAX_PAGE_SIZE } from '@/constants/pagination';
+
+export function useBlockPages(apiUrl: string, address: string) {
+  const maxPagesSafety = MAX_PAGE_SIZE - 1;
+  return useQuery({
+    queryKey: ['block-apges', apiUrl, address],
```
saturn-dbeal: Typo in queryKey: `'block-apges'` → `'block-pages'`

```suggestion
    queryKey: ['block-pages', apiUrl, address],
```

packages/website/src/hooks/useBlockPages.ts
```diff
@@ -0,0 +1,46 @@
...
+      if (!apiUrl) {
+        return { blocks: [], totalTxs: 0 };
+      }
+      const blocksSet = new Set<string>();
+      let isLastPage = false;
+      let block = 0;
+      let iterations = 0;
+      let totalTxs = 0;
+
+      do {
```
saturn-dbeal: **Performance concern:** This fetches ALL pages sequentially on initial load — up to 99 API calls in a `do/while` loop before the user sees anything. For addresses with many transactions (e.g. popular contracts), this could take a very long time and hammer the Otterscan API.

Consider a lazy/on-demand approach instead:
- Only fetch the **current page** data on load
- Fetch the next page boundary when the user clicks "Next"
- Use cursor-based pagination (pass the last block number as the cursor) rather than pre-computing all page boundaries

This would make the initial load O(1) API calls instead of O(n) pages.

packages/website/src/hooks/useBlockPages.ts
```diff
@@ -0,0 +1,46 @@
...
+import { searchTransactions } from '@/lib/address';
+import { MAX_PAGE_SIZE } from '@/constants/pagination';
+
+export function useBlockPages(apiUrl: string, address: string) {
+  const maxPagesSafety = MAX_PAGE_SIZE - 1;
+  return useQuery({
+    queryKey: ['block-apges', apiUrl, address],
+    queryFn: async () => {
+      if (!apiUrl) {
+        return { blocks: [], totalTxs: 0 };
```
saturn-dbeal: The early return uses `blocks` but the successful return uses `pages` — inconsistent key names:
```suggestion
        return { pages: [], totalTxs: 0 };
```

packages/website/src/features/Txs/TransactionsSection.tsx
```diff
@@ -0,0 +1,135 @@
...
+  pageIndex: string | undefined;
+};
+
+const TransactionsSection: React.FC<TransactionsSectionProps> = ({
+  address,
+  chain,
+  pageIndex,
+}) => {
+  const [blockBefore, setBlockBefore] = useState<number | null>(null);
+  const cannonChains = useCannonChains();
```
saturn-dbeal, line 25: Using `useState` + `useEffect` to derive `blockBefore` from `pages` and `numericPageIndex` is an anti-pattern in React — this is derived state that can be computed directly:

```typescript
const blockBefore = useMemo(() => {
  if (!pages || pages.length === 0) return 0;
  return numericPageIndex <= 1 ? 0 : Number(pages[numericPageIndex - 2] ?? '0');
}, [pages, numericPageIndex]);
```

This avoids the extra render cycle caused by the `useEffect` → `setState` pattern.

packages/website/src/features/Txs/TransactionsPagination.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { ChevronRight, ChevronLeft } from 'lucide-react';
+import { MAX_PAGE_SIZE } from '@/constants/pagination';
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from '@/components/ui/tooltip';
+
+type TransactionPagenationProp = {
+  address: string;
```
saturn-dbeal, line 11: Typo in type name: `TransactionPagenationProp` → `TransactionPaginationProp`

packages/website/src/features/Txs/TransactionsPagination.tsx
```diff
@@ -0,0 +1,113 @@
...
+      </Link>
+
+      {/* Prev */}
+      {!isFirstPage ? (
+        <Tooltip>
+          <TooltipTrigger asChild>
+            <Link
+              href={getPageHref(
+                isFirstPage ? 1 : isLastPage ? totalPages - 1 : currentPageIndex
+              )}
```
saturn-dbeal: The "Previous" button href logic is hard to follow:
```typescript
href={getPageHref(isFirstPage ? 1 : isLastPage ? totalPages - 1 : currentPageIndex)}
```

When on the last page, this goes to `totalPages - 1`, but when on a middle page, it goes to `currentPageIndex` (which is 0-indexed). Since `getPageHref` uses 1-indexed pages, this seems like it could be off-by-one. On a middle page (e.g. page 3, `currentPageIndex = 2`), clicking Previous would go to page 2 — that's correct. But the ternary is confusing. Consider simplifying:
```typescript
const prevPage = isLastPage ? totalPages - 1 : currentPageIndex;
href={getPageHref(prevPage)}
```

packages/website/src/features/Txs/TransactionsSection.tsx
```diff
@@ -0,0 +1,135 @@
...
+          </div>
+        </CardContent>
+      </Card>
+    );
+  }
+
+  const { txs, receipts, isLastPage, isFirstPage } = transactionData;
+  const receipt = receipts[receipts.length - 1];
+  const blockNumber = String(parseInt(receipt.blockNumber.slice(2), 16));
+
```
saturn-dbeal, line 131: If `receipts` is an empty array, `receipts[receipts.length - 1]` will be `undefined` and the `parseInt` on the next line will produce `NaN`. Should add a guard:
```typescript
if (!receipts.length) return <EmptyState />;
```

saturn-dbeal (commented):
I've taken over this PR to address the performance concern I flagged earlier.

## Changes Made

### 1. Lazy Loading for Pagination (Performance Fix)
The original implementation pre-fetched ALL page boundaries upfront by looping through all transactions. For addresses with many transactions, this could result in dozens of sequential API calls before anything rendered.

**Solution:** Refactored `useBlockPages` to fetch pages on-demand:
- Pages are discovered as the user navigates
- Only fetches the pages needed for the current navigation
- Caches discovered page boundaries for faster back-navigation

### 2. Fixed Pagination Component
The pagination was using `pages.indexOf(blockNumber)` to determine the current page, which doesn't work with lazy loading since we don't have all pages upfront.

**Solution:** Updated `TransactionsPagination` to:
- Accept explicit `currentPageIndex` and `totalPages` props
- Show `?` for total pages when unknown (haven't reached the end yet)
- Properly disable "Last" button when total pages are unknown

### 3. Fixed Title Typo
`paginatoin` → `pagination`

### 4. Fixed Component Name Typo
`TransactionPagenation` → `TransactionPagination`

---

CI is running to verify everything works. Once checks pass, this should be ready to merge.

PR #1832, opened 2025-09-04
fix(builder): infinite loop possible on cycle in chain definition

dbeal-eth:
due to the way that the call patterns work, it was very possible to have a cycle if dependencies were defined incorrectly.

fixes https://linear.app/usecannon/issue/CAN-763/builder-fails-to-detect-cycle-in-cannonfile


dbeal-eth:
@claude please review

PR #1836, opened 2025-09-09
fix(website): filters to package list

AkihisaY:
Fixed [CAN-746](https://linear.app/usecannon/issue/CAN-746/add-filters-to-package-list)

Screenshot 2025-09-09 at 13 37 47


dbeal-eth (approved):
beautiful 👍

PR #1837, opened 2025-09-18
fix(cli): printing of RPC errors should be debug

dbeal-eth:
due to recent upstream changes RPC errors have become more vocal. likely not necessary to debug log RPC errors anymore as we do a good job of surfacing errors already if they are important


AkihisaY (approved):
Looks good!

PR #1839, opened 2025-10-18
feat(builder): `create2` option support for `router` step

dbeal-eth:
fixes https://linear.app/usecannon/issue/CAN-764/router-does-not-have-create2-option


dbeal-eth:
merging because remaining suggestions are no longer needed

PR #1840, opened 2025-10-18
fix(cli): stall on some forge `cli` builds

dbeal-eth:
When forge outputs large amounts of data (verbose builds, many warnings),
the child process buffers would fill causing the process to block.

This fix actively consumes stdout/stderr streams and displays them
on build failure to improve debuggability.


dbeal-eth:
merging because positive feedback from client, good AI review

PR #1841, opened 2025-10-19
fix(website): clear display when buttons are grayed out due to outsid…

dbeal-eth:
its very confusing when outside safe transactions are grayed out and its unexplained why. This adds a needed error message that may indicate a failure to execute "Outside Safe Transactions" so that its clear the reason why its grayed out.

additionally, there is a logic change/simplification in the button change logic hwere we only unblock the button when there are "incomplete" trnasactions remaining (ie . `deployer.executionProgress.length < deployer.queuedTransactions.length`). This is intended, and is the more correct approach.


dbeal-eth:
merging due to positive feedback from client and decent ai review

PR #1846, opened 2025-11-27
fix(builder): early return from `computeTemplateAccesses`

dbeal-eth:
one of our users indicated that the website deployer was processing the cannonfile extremely slowly.

it turns out that `computeTemplateAccesses`  will use a huge amount of compute to process template accesses, even for strings that have no template usage at all. so to prevent this from being expensive, in most cases we can early return by just doing a simple string check.

### DIFF
--- a/examples/sample-foundry-project/cannonfile.toml
+++ b/examples/sample-foundry-project/cannonfile.toml
@@ -1,6 +1,5 @@
-#:schema https://raw.githubusercontent.com/usecannon/cannon/refs/heads/dev/packages/lsp/src/schema.json
 name = "greeter-foundry"
-version = "1.2.3"
+version = "<%= package.version %>"
 description = "Simple project to verify the functionality of cannon"
 keywords = ["sample", "greeter"]
 
@@ -13,7 +12,6 @@ msg = "my library was deployed in txn <%= contracts.library.deployTxnHash %>"
 [deploy.library]
 artifact = "Library"
 value = "2929372"
-
 
 [deploy.greeter]
 artifact = "Greeter"
--- a/examples/sample-foundry-project/cannonfile.toml
+++ b/examples/sample-foundry-project/cannonfile.toml
@@ -1,5 +1,6 @@
+#:schema https://raw.githubusercontent.com/usecannon/cannon/refs/heads/dev/packages/lsp/src/schema.json
 name = "greeter-foundry"
-version = "<%= package.version %>"
+version = "1.2.3"
 description = "Simple project to verify the functionality of cannon"
 keywords = ["sample", "greeter"]
 
@@ -12,6 +13,7 @@ msg = "my library was deployed in txn <%= contracts.library.deployTxnHash %>"
 [deploy.library]
 artifact = "Library"
 value = "2929372"
+
 
 [deploy.greeter]
 artifact = "Greeter"
--- a/packages/builder/src/definition.ts
+++ b/packages/builder/src/definition.ts
@@ -73,6 +73,8 @@ export class ChainDefinition {
   readonly dependencyFor = new Map<string, string>();
   readonly resolvedDependencies = new Map<string, string[]>();
 
+  readonly templatePossibleNames: string[] = [];
+
   readonly danglingDependencies = new Set<`${string}:${string}`>();
 
   constructor(def: RawChainDefinition, sensitiveDependencies = false) {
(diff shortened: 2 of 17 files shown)


packages/builder/src/definition.ts
```diff
@@ -10,6 +10,7 @@ import { template } from './utils/template';
 
 import { PackageReference } from './package-reference';
 import { ZodIssue } from 'zod';
+import { AccessRecorderEngine } from '.';
```
saturn-dbeal: Importing from the barrel `index.ts` risks a circular dependency since `index.ts` re-exports from `definition.ts` (and other siblings). Safer to import directly:

```suggestion
import { AccessRecorderEngine } from './access-recorder';
```

packages/builder/src/definition.ts
```diff
@@ -73,6 +74,8 @@ export class ChainDefinition {
   readonly dependencyFor = new Map<string, string>();
   readonly resolvedDependencies = new Map<string, string[]>();
 
+  accessRecorderEngine: AccessRecorderEngine|null = null;
```
saturn-dbeal: The `!` non-null assertion on line 260 (`this.accessRecorderEngine!`) is safe in the current call flow (`initializeComputedDependencies` → `checkOutputClash` → `computeDependencies`), but fragile if anyone calls `computeDependencies` independently. Consider a guard:

```typescript
if (!this.accessRecorderEngine) {
  this.computePossibleNames();
}
```

or initialize lazily via a getter.

packages/builder/src/definition.ts
```diff
@@ -449,12 +439,32 @@ export class ChainDefinition {
             outputClashes.push({ output, actions: [this.dependencyFor.get(output)!, fullActionName] });
           }
         }
+
+        this.computePossibleNames();
```
saturn-dbeal: `computePossibleNames()` is called on **every iteration** of the for loop, recreating the `AccessRecorderEngine` each time. Since `dependencyFor` is being built incrementally in the loop, the engine constructed mid-loop will have incomplete possible names anyway. Should be moved **after** the loop so it only runs once with the complete set of names.

examples/sample-foundry-project/cannonfile.toml
```diff
@@ -1,5 +1,6 @@
+#:schema https://raw.githubusercontent.com/usecannon/cannon/refs/heads/dev/packages/lsp/src/schema.json
```
saturn-dbeal: These changes (schema ref, hardcoded version, extra blank line) seem unrelated to the perf fix. Worth splitting into a separate commit or dropping to keep the PR focused.

saturn-dbeal (changes requested):
Verified the performance improvement: **3.3x speedup** on a benchmark with 500 strings (50 steps × 10 config fields). The early `<%=` check alone saves ~40% of template processing calls. Correctness also verified — no state leaks between calls, `clearAccesses()` works properly through the Proxy, and `instanceof` checks are reliable.

However, the **existing tests are broken** — 4 test suites fail (`access-recorder.test.ts`, `invoke.test.ts`, `deploy.test.ts`, `clone.test.ts`) because they still reference the old `computeTemplateAccesses` standalone function and pass `[]` instead of an `AccessRecorderEngine` to `getInputs`. These need to be updated.

Other issues noted in inline comments below.

packages/builder/src/definition.ts
dbeal-eth: guards are a good idea. added.

packages/builder/src/definition.ts
dbeal-eth: I checked and I couldn't find a reason indeed to re-run the compute every iteration of the loop, so it has been moved to after as suggested.

examples/sample-foundry-project/cannonfile.toml
dbeal-eth: undone.

dbeal-eth:
tetss should be fixed. one edge case found while going through your suggested changes/adding the guard is that it is important to check that there are actually actions defined in the cannon definition. If none are defined, then `computePossibleNames` and `checkOutputClash` can potentially loop and stack overflow.

saturn-dbeal (approved):
All previous feedback addressed. Re-verified:

- **Tests:** 21/21 suites passing, 270 tests pass (previously 4 suites were broken)
- **Performance:** 3.3x speedup confirmed on updated code
- **Correctness:** No state leaks between calls, clearAccesses works correctly
- **Import fix:** Now imports from `./access-recorder` directly ✅
- **Guard added:** Lazy init in `computeDependencies` prevents null crash ✅
- **Loop fix:** `computePossibleNames()` moved after the loop ✅
- **Cannonfile changes:** Reverted ✅
- **Stack overflow guard:** The mutual recursion guard in `computePossibleNames` (checking `allActionNames.length > 0 && dependencyFor.size == 0`) is a good catch

LGTM 👍

PR #1848, opened 2026-01-15
fix(builder): prevent flaky/inconsistent rpcs from failing the build

dbeal-eth:
Some RPCs (such as Base or Base Sepolia) utilize large load balancers, and it is possible that after receiving a successfully processed transaction,

As this has become a common problem it seems, an auto-retry has been selectively implemented for the `getBlock` ethereum RPC method, which automatically retries against the RPC with exponential backoff a few times. If the RPC request continues to fail, the original error is returned.

The `promise-retry` module (already a deep dependency of this project) seems well suited to this particular use.

A basic test has been added to verify the new helper function as well.

In the `deploy.ts` action, I did notice that we were querying block by number. It is probably more appropriate to query by blockHash in the small chance that there is a reorg, and it might reduce the chance of flake issues (who knows).

there appears to be a large reduction in the size of our lockflie. I assume this is due to updates or me working on the `update-deps` branch.


packages/builder/src/builder.ts
```diff
@@ -64,6 +63,8 @@ ${printChainDefinitionProblems(problems)}`);
 
   debug('build', initialCtx.settings);
 
+  const { ActionKinds } = await import('./actions');
```
mjlescano, line 67: Why is this PR moving some static imports to async module import?

dbeal-eth: ah yea. I was running into issues with the new tests I added leading to import cycles. Probably this is the import that was causing the trouble https://github.com/usecannon/cannon/pull/1848/changes#diff-5a9ea676f3853e60f82858b2ef44dd6634ac03a46e13dac6f98abf1cab35ada0R11

In the future we will probably want to increase the use of async imports yet further. They provide an effective way to increase code splitting now that we are moving towards ESM for cannon 3.0. For example the website probably does not need to load the execution details of individual actions like `deploy` in most cases.

packages/builder/src/helpers.ts
```diff
@@ -67,3 +68,12 @@ export async function loadPrecompiles(provider: viem.TestClient) {
 
   for (const precompileCall of precompiles.default) await provider.setCode(precompileCall);
 }
+
```
saturn-dbeal: Tiny typos: `hte` → `the`, `propogates` → `propagates`

```suggestion
// due to flaky RPC issues (ex. load balancer where the state propagates unevenly), it is sometimes the case that the block
```

packages/builder/src/helpers.ts
```diff
@@ -67,3 +68,12 @@ export async function loadPrecompiles(provider: viem.TestClient) {
...
+// due to flaky RPC issues (ex. load balancer where hte state propogates unevenly), it is sometimes the case that the block
+// information comes in *after* the receipt is available. This function is intended to provide an auto-retry capability
+// for any errors that could be encountered on this specific problematic call
+export async function getBlockRetried(provider: viem.PublicClient, blockHash: viem.Hash) {
+  return await promiseRetry({ retries: 5, minTimeout: 50 }, (retry) => {
```
saturn-dbeal, line 76: 5 retries with `minTimeout: 50` means worst case ~1.6s total wait (50 + 100 + 200 + 400 + 800). Seems reasonable for load balancer propagation lag. `promise-retry` defaults to `factor: 2` and `maxTimeout: Infinity` — might want to cap `maxTimeout` explicitly if this ever ends up in a hotter path, though in practice 5 retries with factor 2 won't get crazy.

packages/builder/src/builder.ts
saturn-dbeal, line 67: The switch from static `import` to dynamic `await import('./actions')` in three places — is this to break a circular dependency? If so, might be worth a brief comment explaining why, since it's a pattern that could confuse future contributors.

packages/builder/src/steps/deploy.ts
```diff
@@ -408,8 +409,7 @@ const deploySpec = {
       }
     }
 
-    const block = await runtime.provider.getBlock({ blockNumber: receipt?.blockNumber });
-
+    const block = receipt ? await getBlockRetried(runtime.provider, receipt.blockHash) : null;
```
saturn-dbeal, line 412: Good improvement switching from `blockNumber` to `blockHash` — more resilient to reorgs as noted in the PR description. The `receipt` null check is also a nice safety addition.

saturn-dbeal (approved):
Build and all 21 test suites pass (271 tests). The retry logic is clean and the test coverage is good.

A few minor observations (non-blocking):

packages/builder/src/helpers.ts
dbeal-eth: yep. in practice its likely much more than this because of the round trip between the caller and the RPC node

packages/builder/src/builder.ts
dbeal-eth: fixed da1c85a3

PR #1852, opened 2026-01-19
Fix Select Safe dialog overflow

shubham5080:
Screenshot from 2026-01-19 21-33-20


dbeal-eth (approved):
thank you for the contribution!

PR #1854, opened 2026-02-28
feat(cli): add --ipfs-superfluous flag to clean command

saturn-dbeal:
## Summary

Add a new mode to the `cannon clean` command that deletes IPFS packages not referenced by any tag in the tags folder.

## Usage

```bash
# Delete superfluous IPFS packages with confirmation
cannon clean --ipfs-superfluous

# Delete without confirmation (useful for scripts)
cannon clean --ipfs-superfluous --no-confirm
```

## How it works

1. Scans the `tags/` folder for all `.txt` and `.meta` files
2. Extracts IPFS URLs from these files
3. Compares against all files in `ipfs_cache/`
4. Identifies files not referenced by any tag
5. Prompts for confirmation before deletion (unless `--no-confirm`)

## Benefits

- Helps users free disk space by removing orphaned IPFS cache files
- Safe: only removes files that are definitely not referenced
- Shows exactly what will be deleted and how much space will be freed

## Testing

- Unit tests added for the new functionality
- Manually tested with local Cannon cache

---

Draft PR - ready for initial review.

### DIFF
--- a/packages/cli/src/commands/clean.ts
+++ b/packages/cli/src/commands/clean.ts
@@ -1,13 +1,13 @@
 import fs from 'node:fs/promises';
 import { existsSync } from 'node:fs';
 import path from 'node:path';
-import crypto from 'crypto';
 import Debug from 'debug';
 import prompts from 'prompts';
 
 import { log } from '../util/console';
 import { resolveCliSettings } from '../settings';
 import { IPFSLoader } from '@usecannon/builder';
+import { CliLoader } from '../loader';
 
 const debug = Debug('cannon:cli:clean');
 
@@ -65,16 +65,6 @@ export async function clean(confirm = true) {
 }
 
 /**
- * Get the cache file path for an IPFS URL
- * Matches the logic in CliLoader.getCacheHash
- */
-function getCacheHash(url: string): string {
-  const qmhash = url.replace(IPFSLoader.PREFIX, '');
-  const md5 = crypto.createHash('md5').update(qmhash).digest('hex');
-  return `${md5}-${qmhash.toLowerCase()}`;
-}
-
-/**
  * Read all tag files and extract referenced IPFS URLs
  */
 async function getReferencedIpfsUrls(tagsDir: string): Promise<Set<string>> {
@@ -129,7 +119,7 @@ async function resolveIpfsDependencies(rootUrls: Set<string>, ipfsCacheDir: stri
 
   while (queue.length > 0) {
     const url = queue.pop()!;
-    const cacheHash = getCacheHash(url);
+    const cacheHash = CliLoader.getCacheHash(url);
     const cacheFile = path.join(ipfsCacheDir, `${cacheHash}.json`);
 
     try {
@@ -167,16 +157,15 @@ async function getIpfsCacheFiles(ipfsCacheDir: string): Promise<string[]> {
 export interface CleanIpfsStats {
   totalFiles: number;
   referencedFiles: number;
-  superfluousFiles: number;
+  orphanedFiles: number;
   deletedFiles: number;
   failedFiles: number;
-  freedBytes: number;
 }
 
 /**
- * Clean superfluous IPFS packages (packages not referenced by any tag)
+ * Clean orphaned IPFS packages (packages not referenced by any tag)
  */
-export async function cleanSuperfluousIpfs(confirm = true): Promise<{ success: boolean; stats: CleanIpfsStats }> {
+export async function cleanOrphanedIpfs(confirm = true): Promise<{ success: boolean; stats: CleanIpfsStats }> {
   const settings = resolveCliSettings();
 
   const tagsDir = path.join(settings.cannonDirectory, 'tags');
@@ -185,10 +174,9 @@ export async function cleanSuperfluousIpfs(confirm = true): Promise<{ success: b
   const stats: CleanIpfsStats = {
     totalFiles: 0,
     referencedFiles: 0,
-    superfluousFiles: 0,
+    orphanedFiles: 0,
     deletedFiles: 0,
     failedFiles: 0,
-    freedBytes: 0,
   };
 
   // Get all directly referenced IPFS URLs from tags
@@ -207,11 +195,11 @@ export async function cleanSuperfluousIpfs(confirm = true): Promise<{ success: b
   // Build a set of referenced cache hashes for quick lookup
   const referencedHashes = new Set<string>();
   for (const url of referencedUrls) {
-    referencedHashes.add(getCacheHash(url));
+    referencedHashes.add(CliLoader.getCacheHash(url));
   }
 
-  // Find superfluous files (not referenced by any tag)
-  const superfluousFiles: { file: string; size: number }[] = [];
+  // Find orphaned files (not referenced by any tag)
+  const orphanedFiles: string[] = [];
 
   for (const file of cacheFiles) {
     // Extract the hash part from the filename (format: {md5}-{cid}.json)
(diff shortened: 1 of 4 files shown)


packages/cli/src/commands/config/index.ts
```diff
@@ -795,6 +795,10 @@ export const commandsConfig: CommandsConfig = {
         flags: '--no-confirm',
         description: 'Do not ask for confirmation before deleting',
       },
+      {
+        flags: '--ipfs-superfluous',
```
dbeal-eth: lets rename this flag to just `--ipfs`. The idea is that we are removing any IPFS artifacts that are unneeded. When the flag is not specified, it removes literally everything.

packages/cli/src/commands/clean.ts
```diff
@@ -61,3 +63,181 @@ export async function clean(confirm = true) {
...
+
+/**
+ * Get the cache file path for an IPFS URL
+ * Matches the logic in CliLoader.getCacheHash
+ */
+function getCacheHash(url: string): string {
+  const qmhash = url.replace(IPFSLoader.PREFIX, '');
+  const md5 = crypto.createHash('md5').update(qmhash).digest('hex');
+  return `${md5}-${qmhash.toLowerCase()}`;
+}
```
dbeal-eth: in the comment you say we are matching the logic from elsewhere. Can we refactor this function form the source rather than creating a duplicate one here?

packages/cli/src/commands/clean.ts
```diff
@@ -61,3 +63,181 @@ export async function clean(confirm = true) {
...
+  const referencedUrls = await getReferencedIpfsUrls(tagsDir);
+  debug(`Found ${referencedUrls.size} referenced IPFS URLs in tags`);
+
+  // Get all IPFS cache files
+  const cacheFiles = await getIpfsCacheFiles(ipfsCacheDir);
+  stats.totalFiles = cacheFiles.length;
+  debug(`Found ${cacheFiles.length} IPFS cache files`);
+
+  // Build a set of referenced cache hashes for quick lookup
+  const referencedHashes = new Set<string>();
```
dbeal-eth, line 195: Can we instead have the `getReferencedIpfsUrls` return a set directly? I don't believe its going to be used anywhere else for the forseeable future.

packages/cli/src/commands/clean.ts
```diff
@@ -61,3 +63,181 @@ export async function clean(confirm = true) {
...
+    } catch (error: unknown) {
+      stats.failedFiles++;
+      if (error instanceof Error) {
+        debug(`Failed to delete ${file}: ${error.message}`);
+      }
+    }
+  }
+
+  log();
+  log(`Deleted ${stats.deletedFiles} superfluous IPFS package(s), freed ${formatBytes(stats.freedBytes)}.`);
```
dbeal-eth: instead of using the terminology "superfluous", we should probably just say "orphaned" or "unreferenced".

packages/cli/src/commands/clean.ts
```diff
@@ -61,3 +63,181 @@ export async function clean(confirm = true) {
...
+    log(`Failed to delete ${stats.failedFiles} file(s).`);
+  }
+
+  return { success: true, stats };
+}
+
+/**
+ * Format bytes to human-readable string
+ */
+function formatBytes(bytes: number): string {
```
dbeal-eth: lets go ahead and remove this function, and remove all the functionality that calculates how much space was freed. I don't think its needed (let me know if you think I am mistaken in thinking this)

dbeal-eth (changes requested):
This is looking good so far. I particularly like how you used careful validation throughout the scanning processes.

one thing we have to be careful about is that its possible for an IPFS file to contain a dependency upon another IPFS file. In this case, we should be keeping those second-order IPFS files as well. Please make changes to ensure this isn't an issue 👍

saturn-dbeal:
Addressed review feedback: added recursive IPFS dependency resolution. The clean command now:
1. Finds directly referenced IPFS URLs from tags
2. Reads the content of each cached IPFS file and extracts nested `ipfs://` URLs
3. Recursively follows the dependency chain until all transitive dependencies are found
4. Only marks files as superfluous if they're not in the complete dependency graph

Added tests covering the second-order dependency case.

packages/cli/src/commands/config/index.ts
saturn-dbeal: Renamed to `--ipfs` and updated the description.

packages/cli/src/commands/clean.ts
saturn-dbeal: Removed the duplicate and imported `CliLoader.getCacheHash` directly from `../loader`.

packages/cli/src/commands/clean.ts
saturn-dbeal: Renamed all references from "superfluous" to "orphaned" throughout — variable names, function names, log messages, and interface fields.

packages/cli/src/commands/config/index.ts
saturn-dbeal: Done — renamed to `--ipfs` with updated description: "Delete only orphaned IPFS packages not referenced by any tag".

saturn-dbeal:
All inline review comments addressed in 7917b33. Changes:
- Renamed `--ipfs-superfluous` → `--ipfs`
- Removed duplicate `getCacheHash`, now imports `CliLoader.getCacheHash`
- Confirmed `getReferencedIpfsUrls` already returns `Set<string>`
- Renamed all "superfluous" → "orphaned" terminology
- Removed `formatBytes` and space-freed calculations

packages/cli/src/commands/clean.ts
saturn-dbeal: Done — removed `formatBytes`, `freedBytes`, and all file size tracking. Output now just reports count of deleted files.

saturn-dbeal:
All inline review comments have been addressed. Ready for re-review. Changes in commit e2a6600:

- Renamed `--ipfs-superfluous` → `--ipfs`
- Removed duplicate `getCacheHash`, now imports `CliLoader.getCacheHash`
- Confirmed `getReferencedIpfsUrls` already returns `Set<string>`
- Renamed all "superfluous" terminology to "orphaned"
- Removed `formatBytes` and all space-freed tracking

packages/cli/src/commands/config/index.ts
saturn-dbeal: Renamed to `--ipfs` and updated the description.

packages/cli/src/commands/clean.ts
saturn-dbeal: Refactored to import `getCacheHash` from CliLoader instead of duplicating the logic.

packages/cli/src/commands/clean.ts
saturn-dbeal: Confirmed — it already returns `Set<string>` directly, no changes needed.

packages/cli/src/commands/clean.ts
saturn-dbeal: Renamed all "superfluous" references to "orphaned" throughout — function names, variable names, interface fields, and log messages.

packages/cli/src/commands/clean.ts
saturn-dbeal: Removed `formatBytes` and all space-freed tracking. Output is simplified now.

dbeal-eth:
e2e tests are not going to pass here due to lacking credentials. Going to merge as is.

PR #1855, opened 2026-02-28
feat(cli): cache all package tags from registry to local tags folder

saturn-dbeal:
## Summary

Previously, only immutable packages (`mutability === 'version'`) were cached locally when fetched from the on-chain registry. This change caches ALL package references, including mutable tags.

## Benefits

- **Faster subsequent lookups** for previously fetched packages
- **Offline access** to cached packages
- **Consistent caching behavior** with IPFS artifacts (which are cached on download)

## Changes

- Remove mutability check in `createDefaultReadRegistry` event handler
- Also cache metadata URL when available
- Preserve mutability information in cached entries
- Add unit tests for tag caching functionality

## How it works

When a package is fetched from the on-chain registry, its reference (IPFS URL) is now automatically saved to the local `tags/` folder, similar to how IPFS artifacts are cached. This means:

1. First fetch: resolves from on-chain registry, caches locally
2. Subsequent fetches: can resolve from local cache (faster, works offline)

---

Draft PR - ready for initial review.

### DIFF
--- a/packages/cli/src/registry.test.ts
+++ b/packages/cli/src/registry.test.ts
@@ -1,0 +1,104 @@
+import { LocalRegistry } from './registry';
+import { FallbackRegistry } from '@usecannon/builder';
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import os from 'node:os';
+
+describe('LocalRegistry tag caching', () => {
+  let tempDir: string;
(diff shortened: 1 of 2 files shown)


saturn-dbeal:
Closing in favor of a new PR with the correct fix (hexToString on mutability field + corrected publish call signature).

PR #1857, opened 2026-03-01
feat(lsp): add cannonfile fragment schema and include field

saturn-dbeal:
## Summary

This PR adds support for **cannonfile fragments** - TOML files that are meant to be included from a main cannonfile and don't require top-level metadata like `name` and `version`.

### 1. Add `include` field to `chainDefinitionSchema`
```toml
include = [
    "tomls/settings.toml",
    "tomls/core.toml",
    "tomls/permissions.toml"
]
```
This field was already parsed by the CLI but was missing from the schema, causing IDE warnings.

### 2. Add `cannonfileFragmentSchema`
A new schema for included fragments that omits:
- `name` (required for publishing)
- `version` (required for publishing)
- `preset` (optional in both)
- `include` (only main cannonfile should include)
- `privateSourceCode` (only main cannonfile should define)
- `description` (only main cannonfile should define)
- `keywords` (only main cannonfile should define)
- `deployers` (only main cannonfile should define)

Fragments can contain all the same actions: `deploy`, `invoke`, `clone`, `pull`, `var`, `router`, `diamond`, etc.

### 3. Generate both schemas
- `schema.json` - Full cannonfile schema
- `schema-fragment.json` - Fragment schema for included files

## Usage

IDEs/LSPs can now provide proper validation for:
- **Full cannonfiles**: Must have `name` and `version`, can have `include`
- **Fragment files**: Only actions, no metadata required

## Example Fragment (`tomls/core.toml`)

```toml
[var.system_config]
core_package = "my-protocol-core:1.0.0"

[deploy.Core]
artifact = "Core"
args = ["<%= settings.owner %>"]
```

This fragment can be included from a main cannonfile without causing validation errors about missing `name`/`version`.

---

Draft PR - ready for review.


saturn-dbeal:
Updated to also include the `include` field in the fragment schema, so fragments can include other fragments.

dbeal-eth (approved):

packages/builder/src/schemas.ts
```diff
@@ -1044,3 +1049,104 @@ export const chainDefinitionSchema = z
...
+ * Fragments can contain all the same actions as a full cannonfile but omit:
+ * - name (required for publishing)
+ * - version (required for publishing)
+ * - preset (optional in both)
+ * - privateSourceCode (only main cannonfile should define)
+ * - description (only main cannonfile should define)
+ * - keywords (only main cannonfile should define)
+ * - deployers (only main cannonfile should define)
+ */
+export const cannonfileFragmentSchema = z
```
dbeal-eth, line 1071: there is a lot of code duplication here. Since the full schema is just the fragment schema , there has got to be some way that we can deduplicate this, right?

dbeal-eth (changes requested):

saturn-dbeal:
Good call — extracted the shared actions into `cannonfileActionsSchema` and both the full schema and fragment schema now compose from it. ~76 lines removed.

saturn-dbeal:
Addressed review feedback: deduplicated the schemas by extracting a shared `cannonfileActionsSchema` that both the full schema and fragment schema extend from. No more code duplication.

packages/builder/src/schemas.ts
saturn-dbeal: Done — refactored so the fragment schema is defined first, then the full `chainDefinitionSchema` extends it by adding the top-level package fields (`name`, `version`, `preset`, etc.). No more duplication.

saturn-dbeal:
Inline review comment addressed in 66411d0 — deduplicated the fragment and full schemas so the full schema extends the fragment.

dbeal-eth:
tests will not pass due to PR being contributed from external, so merging with bypassed rues.

dbeal-eth (approved):

PR #1858, opened 2026-03-01
fix(cli): prevent spinner output corruption

saturn-dbeal:
## Summary

Fixes two issues with spinner output corruption in the CLI.

## Issue 1: Builder package uses console.log directly (HIGH PRIORITY)

**Problem:** The builder package (`registry.ts`) logs directly with `console.log`, which doesn't coordinate with the CLI's spinner. This causes output corruption when running commands like `cannon publish`, `cannon prune`, etc.

**Before (corrupted):**
```
⠋ Fetching...Package: my-package  <<<< CORRUPTED: spinner and text overlap on same line
Preset: main
Tags: 1.0.0, latest
```

The first line shows the corruption: `console.log` writes "Package: my-package" while the spinner is still active, causing them to overlap on the same line.

**After (fixed):**
```
⠋ Fetching...
Package: my-package
Preset: main
Tags: 1.0.0, latest
```

**Fix:**
- Extracted `Logger` interface to its own module (`packages/builder/src/logger.ts`)
- CLI provides a spinner-aware logger via `setBuilderLogger()`
- All `registry.ts` `console.log` calls now use the injected logger
- Added eslint rule to disallow `console.log` in builder package

## Issue 2: Sequential logSpinner calls cause flickering (MEDIUM PRIORITY)

**Problem:** Commands like `cannon prune` call `logSpinner()` multiple times in sequence. Each call stops/starts the spinner, causing visible flickering.

**Fix:** Stop spinner once, output all lines with `log()`, no flickering.

## Changes

| File | Change |
|------|--------|
| `packages/builder/src/logger.ts` | New `Logger` interface and `setBuilderLogger`/`getBuilderLogger` |
| `packages/builder/src/registry.ts` | Use injected logger instead of `console.log` |
| `packages/builder/src/index.ts` | Export `setBuilderLogger` and `Logger` type |
| `packages/cli/src/util/console.ts` | Call `setBuilderLogger` directly at import |
| `packages/cli/src/index.ts` | Fix `prune` output |


packages/builder/src/registry.ts
```diff
@@ -10,9 +10,21 @@ import { prepareMulticall, TxData } from './multicall';
 import { PackageReference } from './package-reference';
 import { CannonSigner } from './types';
 import { getIpfsUrl } from './ipfs';
+import { Logger, defaultLogger } from './logger';
 
 const debug = Debug('cannon:builder:registry');
 
+// Module-level logger that can be set by CLI to coordinate with spinner
```
dbeal-eth: Lets raise this logging capability up to its own module. It would be useful to have this capability elsewhere in the builder for the future.

packages/cli/src/util/console.ts
```diff
@@ -111,3 +112,12 @@ export const logSpinnerEnd = () => {
     // Ignore spinner errors on cleanup
   }
 };
+
+/**
+ * Logger for builder package that coordinates with the CLI spinner
+ */
+export const builderLogger: Logger = {
```
dbeal-eth, line 119: why not just call `setBuilderLogger` directly in this file?

dbeal-eth (changes requested):
the example that you gave of it "fixed" in the PR description seems incorrect. The corruption is in the first line that you pasted:

```
⠋ Fetching...Package: my-package <<<<< THIS IS THE CORRUPTED TEXT
Preset: main
```

(I do think that the change you made instaed was also good, so you can go ahead and keep that 👍 )

Also, lease update the `eslint` configuration (its inline in the file, right?) to disallow `console.log`, since it appears we no longer want to be using it in the builder.

saturn-dbeal:
Addressed all feedback:
- Moved `setBuilderLogger` and logger state into the dedicated `logger.ts` module (with `getBuilderLogger()` for access)
- `console.ts` now calls `setBuilderLogger` directly
- Removed `eslint-disable no-console` from `registry.ts` (no-console is already enforced project-wide)

saturn-dbeal:
Addressed all review feedback:
- Extracted logging into its own module (`packages/builder/src/logger.ts`) with `Logger` interface, `setBuilderLogger`, and `getBuilderLogger`
- `console.ts` now calls `setBuilderLogger` directly
- Added eslint-disable comments for the intentional console usage in the default logger (the global `no-console` rule already existed)
- Note: the PR description example will need to be updated when merging — the corruption is in the first line as you noted

packages/builder/src/registry.ts
saturn-dbeal: Done — extracted the logger to its own module at `packages/builder/src/logger.ts`. It exports the `Logger` interface, `defaultLogger`, and `setBuilderLogger`/`getBuilderLogger`. The registry now imports from `./logger` instead of defining it inline.

packages/cli/src/util/console.ts
saturn-dbeal: Done — `console.ts` now imports and calls `setBuilderLogger` directly from `@usecannon/builder` instead of going through the intermediate `builderLogger` object.

saturn-dbeal:
All inline review comments addressed:
- Extracted logger to its own module (`packages/builder/src/logger.ts`)
- `console.ts` now calls `setBuilderLogger` directly
- Added eslint rule to disallow `console.log` in builder

dbeal-eth (approved):

PR #1868, opened 2026-03-20
fix(clone): republish when built state equals previous state

saturn-dbeal:
Edge case fix: when the built state is exactly equal to the previous state, we skip generating a new deploy URL. However, the reference to the deployed package might not be recorded in the registry. This adds a republish call to ensure the package reference is properly recorded.


dbeal-eth (approved):
looks good, but would be best if we could have a test specifically validating this case

PR #1869, opened 2026-03-31
feat(builder): add safe_set_signers action for Safe multisig owner management

saturn-dbeal:
Adds a new `safe_set_signers` Cannon action that automatically configures the signer list on a Gnosis Safe multisig.

## What it does

Given a Safe address and a desired list of signers + threshold, this action:
1. Reads current owners and threshold from the Safe
2. Compares to the desired configuration
3. Computes the minimal set of Safe calls (add/remove/swap owners, change threshold)
4. Batches all calls into a single `execTransaction`
5. Idempotent — no-op if current signers already match desired

## Files changed

- `packages/builder/src/steps/safe-set-signers.ts` — action implementation (363 lines)
- `packages/builder/src/steps/safe-set-signers.test.ts` — unit tests (228 lines)
- `packages/builder/src/schemas.ts` — zod schema + cannonfile registration
- `packages/builder/src/actions.ts` — action registration

## Usage example (cannonfile)

```toml
[setting.safeAddress]
  value = "0x..."  # Safe contract address

[safe_set_signers.configure-safe]
  target = "${safeAddress}"
  signers = ["0xOwner1", "0xOwner2", "0xOwner3"]
  threshold = 2
```

### DIFF
--- a/packages/builder/src/actions.ts
+++ b/packages/builder/src/actions.ts
@@ -11,6 +11,7 @@ import pullSpec from './steps/pull';
 import routerSpec from './steps/router';
 import diamondSpec from './steps/diamond';
 import varSpec from './steps/var';
+import safeSetSignersSpec from './steps/safe-set-signers';
 import { ChainArtifacts, ChainBuilderContext, PackageState } from './types';
 
 export interface RawConfig {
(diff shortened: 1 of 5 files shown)


dbeal-eth:
as much as I like the idea of this feature, going to cancel it because I think this probably qualifies as out of scope, at least for now. Maybe revisit making this as a plugin in the near future.

PR #1870, opened 2026-04-11
fix(skill): declare environment variables in SKILL.md frontmatter

saturn-dbeal:
## Problem
ClawHub flags the Cannon skill as suspicious because the SKILL.md references environment variables (CANNON_PRIVATE_KEY, RPC_URL, Etherscan API key) but the frontmatter doesn't declare them. This creates a transparency mismatch — users and agents can't see what secrets the skill needs without reading the entire file.

## Changes
Added `env` section to the SKILL.md frontmatter declaring:
- **CANNON_PRIVATE_KEY** (sensitive) — for signing on-chain transactions
- **RPC_URL** — for target chain RPC endpoints
- **ETHERSCAN_API_KEY** (sensitive) — for contract verification
- **CANNON_DIRECTORY** — custom storage path override

All are marked `required: false` since local development (chain 13370, dry-run) doesn't need them.

## Testing
- Frontmatter is valid YAML
- No changes to skill body or behavior

### DIFF
--- a/packages/skill/SKILL.md
+++ b/packages/skill/SKILL.md
@@ -1,6 +1,23 @@
 ---
 name: cannon
 description: Cannon package manager for Ethereum deployments. Use when building, testing, deploying, or inspecting Cannon packages. Covers cannonfile syntax, CLI commands (build, run, test, publish, inspect), actions (deploy, invoke, clone, pull, router, diamond), and package workflows. NOT for general Solidity development questions.
+env:
+  CANNON_PRIVATE_KEY:
+    description: Private key for signing on-chain transactions (deploy, publish, register). Required for non-dry-run operations on real networks.
+    required: false
+    sensitive: true
+  RPC_URL:
+    description: Ethereum RPC endpoint URL for target chain. Required for non-local deployments and publish operations.
+    required: false
+    sensitive: false
+  ETHERSCAN_API_KEY:
+    description: Etherscan API key for contract verification (cannon verify). Required for verifying deployed contracts on Etherscan-supported chains.
+    required: false
+    sensitive: true
+  CANNON_DIRECTORY:
+    description: Custom directory for Cannon package storage and build artifacts. Defaults to ~/.local/share/cannon/ if not set.
+    required: false
+    sensitive: false
 ---
 
 # Cannon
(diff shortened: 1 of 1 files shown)


packages/skill/SKILL.md
```diff
@@ -1,6 +1,23 @@
 ---
 name: cannon
 description: Cannon package manager for Ethereum deployments. Use when building, testing, deploying, or inspecting Cannon packages. Covers cannonfile syntax, CLI commands (build, run, test, publish, inspect), actions (deploy, invoke, clone, pull, router, diamond), and package workflows. NOT for general Solidity development questions.
+env:
```
dbeal-eth: I did a bit of research, and it doesn't seem like we should be using the YAML Frontmatter to define the sensitive environment variables. Instead, we should just add it to the regular skill text body, possibly in a new section deep diving into environment variables.

packages/skill/SKILL.md
```diff
@@ -1,6 +1,23 @@
...
+  CANNON_PRIVATE_KEY:
+    description: Private key for signing on-chain transactions (deploy, publish, register). Required for non-dry-run operations on real networks.
+    required: false
+    sensitive: true
+  RPC_URL:
```
dbeal-eth: its actually `CANNON_RPC_URL`. Make sure to update the envvar references elsewhere if need be.

packages/skill/SKILL.md
```diff
@@ -1,6 +1,23 @@
...
+    description: Ethereum RPC endpoint URL for target chain. Required for non-local deployments and publish operations.
+    required: false
+    sensitive: false
+  ETHERSCAN_API_KEY:
```
dbeal-eth: also has a `CANNON_` envvar prefix. maybe good to just mention in general that all the envvars that can be recognized by cannon are prefixed with `CANNON_`

dbeal-eth (changes requested):
minor things.

packages/skill/SKILL.md
saturn-dbeal: Good point. Moved all env var documentation out of the frontmatter into a new `## Environment Variables` section in the body with a table format.

packages/skill/SKILL.md
saturn-dbeal: Fixed. Renamed to `CANNON_RPC_URL`.

packages/skill/SKILL.md
saturn-dbeal: Added a note in the new Environment Variables section: "All Cannon-specific environment variables use the `CANNON_` prefix."

dbeal-eth (approved):

PR #1871, opened 2026-04-11
fix(verify): etherscan v2 API compatibility and bug fixes

saturn-dbeal:
## Problem

Etherscan v1 API was deprecated August 2025. The cannon verify command needs updates for v2 compatibility, plus several bugs in the verification flow.

### Etherscan v2 API compatibility
- Move `chainid` from POST body to query string parameter for `verifysourcecode` and `checkverifystatus` endpoints (v2 requirement)
- `isEtherscanVerified` already uses query params via URLSearchParams, so no change needed there

### Bug fixes
- **Delete `compilationTarget`** from standard JSON input before sending to Etherscan. Without this, the `compilationTarget` in settings conflicts with the top-level `contractname` parameter, causing verification failures.
- **Fix `checkverifystatus` guid reference** — was passing `guids[c]` (the whole sub-object) instead of `guids[c][etherscan]` (the actual guid string)
- **Use `_.set` for etherscan guid assignment** — direct assignment `guids[c][etherscan]` crashes when `guids[c]` is undefined

## Testing
- Built locally, no TS compilation errors from changes
- Tested against real Etherscan v2 API for contract verification

### DIFF
--- a/packages/cli/src/commands/verify.ts
+++ b/packages/cli/src/commands/verify.ts
@@ -109,6 +109,7 @@ export async function verify(
 
       // supply any linked libraries within the inputs since those are calculated at runtime
       const inputData = JSON.parse(contractArtifact.source.input);
+      delete inputData.settings.compilationTarget;
       _.set(inputData, 'settings.libraries', contractInfo.linkedLibraries);
 
       if (service === 'etherscan' || service === 'all') {
@@ -118,7 +119,6 @@ export async function verify(
           try {
             const reqData: { [k: string]: string } = {
               apikey: cliSettings.etherscanApiKey,
-              chainid: chainId.toString(),
               module: 'contract',
               action: 'verifysourcecode',
               contractaddress: contractInfo.address,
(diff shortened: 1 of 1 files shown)


packages/cli/src/commands/verify.ts
```diff
@@ -118,7 +119,6 @@ export async function verify(
           try {
             const reqData: { [k: string]: string } = {
               apikey: cliSettings.etherscanApiKey,
-              chainid: chainId.toString(),
```
dbeal-eth, line 121: i think we should leave this here--im not sure but it may literally be required to specify *BOTH* in the querystring and the post data. I know for sure leaving it here doesn't hurt anything, at least.

packages/cli/src/commands/verify.ts
```diff
@@ -223,13 +227,17 @@ export async function verify(
     for (const v in guids[c]) {
       for (;;) {
         if (v === 'etherscan') {
+          // Etherscan v2 requires chainid as a query string parameter
+          const etherscanStatusUrl = new URL(etherscanApi);
+          etherscanStatusUrl.searchParams.set('chainid', chainId.toString());
```
dbeal-eth: since this URL is the same as the one for `etherscanVerifyUrl` earlier, can we just unify this into `etherscanUrl` up above?

dbeal-eth (changes requested):

packages/cli/src/commands/verify.ts
saturn-dbeal: Good call. Added `chainid` back to the POST body data alongside the query string param. Kept both to be safe.

packages/cli/src/commands/verify.ts
saturn-dbeal: Done. Replaced both `etherscanVerifyUrl` and `etherscanStatusUrl` with a single `etherscanUrl` computed once after `etherscanApi` is defined.

dbeal-eth (approved):

PR #1872, opened 2026-04-12
feat(alter): populate missing steps for incomplete packages

saturn-dbeal:
## Summary

Allows `cannon alter` to work on packages that have not been fully generated, gated behind a `--populate-missing` flag.

## Problem

It was impossible to alter a cannon package that had not been fully generated/built. Missing subpackages or step definitions would cause errors, blocking workflows that needed to import or modify partial deployments.

### New flag: `--populate-missing`

All new behaviors are gated behind this flag. Without it, the default strict behavior is preserved (typos and missing steps throw errors).

### With `--populate-missing` enabled:

**Subpackage auto-population:**
- Missing subpkg steps are auto-populated with a SKIP hash instead of throwing
- Allows traversing into subpackages that were never built

**Import command improvements:**
- `getConfig` and `getPackageRef` are wrapped in try/catch to handle missing step definitions
- When a step is not in the def, a minimal config is synthesized from transaction receipts (targets + empty ABI)
- Imported artifacts for unresolved steps are recorded with a SKIP hash

### Unconditional fixes:
- `restoreMisc` calls are now guarded with `miscUrl` existence check (prevents errors on packages without misc data)

## Test Plan
- Test `alter import` on a package with missing step definitions using `--populate-missing`
- Test `alter` with subpackages using `--populate-missing`
- Verify default behavior without flag still throws on missing steps/subpkgs
- Test `restoreMisc` guard on packages without miscUrl

### DIFF
--- a/packages/cli/src/commands/alter.ts
+++ b/packages/cli/src/commands/alter.ts
@@ -89,9 +89,7 @@ export async function alter(
     if (!parentDeployInfo?.state[pathItem]) {
       if (!populateMissing) {
         throw new Error(
-          'subpkg path name not found: ' +
-            pathItem +
-            '. Use --populate-missing to auto-populate missing subpackages.'
+          'subpkg path name not found: ' + pathItem + '. Use --populate-missing to auto-populate missing subpackages.'
         );
       }
       debug('auto-populating missing subpkg', pathItem);
@@ -101,7 +99,7 @@ export async function alter(
         hash: 'SKIP' as const,
         version: BUILD_VERSION,
       };
-      startDeployInfo.push({ def: { name: '', version: '' }, state: {}, options: {} });
+      startDeployInfo.push({ def: { name: '', version: '' }, state: {}, options: {} } as DeploymentInfo);
     } else {
       startDeployInfo.push(
         await runtime.readBlob(parentDeployInfo.state[pathItem].artifacts.imports![pathItem.split('.')[1]].url)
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/alter.ts
```diff
@@ -84,12 +85,26 @@ export async function alter(
...
+        );
+      }
+      debug('auto-populating missing subpkg', pathItem);
+      const subpkgName = pathItem.split('.')[1];
+      parentDeployInfo!.state[pathItem] = {
+        artifacts: { contracts: {}, txns: {}, extras: {}, imports: { [subpkgName]: { url: '' } } },
+        hash: 'SKIP' as const,
+        version: BUILD_VERSION,
+      };
+      startDeployInfo.push({ def: { name: '', version: '' }, state: {}, options: {} } as DeploymentInfo);
```
dbeal-eth: lets avoid just casting here. instead, can we populate any missing fields or fill any requirements of the type? if its extremely inconvenient/impossible to do this, please explain why.

dbeal-eth (changes requested):

packages/cli/src/commands/alter.ts
saturn-dbeal: Good point. Populated all required DeploymentInfo fields instead: generator, timestamp, def (with valid name/version), state, options, meta, and miscUrl. The placeholder name is `'placeholder'` and version `'0.0.0'` since this object is never actually used — it just needs to satisfy the type so the pop logic works correctly.

dbeal-eth (approved):

PR #1876, opened 2026-05-30
fix: prevent 'nonce too low' on cannon build -- broadcast (strict-ordering RPCs)

### LINKED ISSUE #1875: cannon build ... -- broadcast produces "nonce too low" errors via viem on strict-ordering RPCs (parallel tx submission)
### Summary

When `cannon build --provider-url <real_rpc> --private-key <pk> -- broadcast`
runs against a real chain, transactions are submitted in parallel/rapid
succession faster than the next-nonce can be re-fetched, producing repeated
`InvalidTransactionNonce` errors from viem like:

```
Details: nonce too low: next nonce 369320, tx nonce 369319
Version: viem@2.30.0
```

The build aborts mid-way; you have to re-run `cannon build --upgrade-from` to
pick up where it stopped. On a large omnibus (e.g. synthetix v3 with ~280
steps) this can take **4+ retries** before all steps complete.

Cannon version: `2.25.1`
viem version (transitive): `2.30.0`
Target chain: MegaETH testnet (chainId 6343), single deployer key with no
external txs competing for the nonce.

### Evidence

Across 4 sequential retries of the same omnibus build:

| Run | "nonce too low" errors | Step skips |
| --- | --- | --- |
| Retry 1 | 4 | 93 |
| Retry 2 | 4 | 20 |
| Retry 3 | 2 | 3 |
| Retry 4 | 0 | 0 |

Sample failure context:

```
nonce:                 369319

Details: nonce too low: next nonce 369320, tx nonce 369319
Version: viem@2.30.0)
  Executing [invoke.enable_feature_claimRewards]...
    ✔ Successfully called setFeatureFlagAllowAll(...)
```

The deployer key was idle outside cannon's invocation, so nonce `369320`
arriving "first" can only mean cannon already submitted that tx while a
sibling step was still using `369319`.

### Expected behavior

`cannon build ... -- broadcast` should serialize tx submission per signer, or
re-fetch the current nonce from the RPC before each submit when running in
real-broadcast mode. `--dry-run` against anvil doesn't show this because
anvil's nonce handling is more lenient.

### Suggested mitigations

1. Serialize broadcasts per `(signer, chainId)` — wait for `eth_sendRawTransaction`
   to return (or for receipt) before submitting the next tx from the same key.
2. As a less invasive option: re-`eth_getTransactionCount(addr, "pending")`
   immediately before signing each tx, instead of relying on a cached
   counter.
3. Document the recommended remediation (`cannon build --upgrade-from`
   loop until 0 skips) so users don't think the deploy is "stuck".

### Impact

In our case the deploy eventually completed cleanly across 4 retries — final
onchain state matches cannon's local state, no data corruption. The
operational impact was the extra hour of retry loops + a misleading
"router seems missing selectors" red herring that took further investigation
to rule out (turned out unrelated, but the nonce-skew failure mode made it
look plausible).

### Repro

Not easy to reproduce on a fast public RPC because most chains tolerate
out-of-order nonces in the mempool. MegaETH's RPC enforces strict nonce
ordering on `eth_sendRawTransaction`, which made the issue surface. Any
chain with similar strict ordering should exhibit it.

alxwlw:
## Summary

Fixes #1875.

`cannon build ... -- broadcast` aborts with `nonce too low: next nonce N+1, tx nonce N` (viem `InvalidTransactionNonce`) on strict-ordering / eventually-consistent RPCs (observed on MegaETH testnet, chainId 6343).

### Root cause (corrected)

The build is strictly sequential — there is no parallel step execution. The real cause is that live signers built from `--private-key` (`packages/cli/src/util/provider.ts`) had **no viem nonce manager**, so every broadcast relied on a fresh `eth_getTransactionCount(addr, 'pending')` read via `prepareTransactionRequest`. When that read is served by a lagging replica it can return a **stale, too-low** nonce after the previous transaction already consumed it → `nonce too low`. This also explains why repeated `--upgrade-from` retries gradually converge.

### Fix

- **CLI:** attach an explicit per-signer viem nonce manager (`createNonceManager({ source: jsonRpc() })`) to all live private-key signers (new `createPrivateKeySigner` helper). Nonce assignment is serialized per `(address, chainId)` and never regresses below an already-used nonce; the manager's `jsonRpc` source reads `eth_getTransactionCount` through the same transport the transactions are sent on.
- **Builder:** wrap every live broadcast — `invoke` / `deploy` / `router` / `diamond` / `create2` and the `onchain-store` upgrade-info writes (`writeUpgradeFromInfo`) — in `sendTransactionWithRetry`: any failed broadcast is retried with bounded backoff, always assuming the nonce may be broken (per review — no error-string matching). Before each retry the signer's nonce manager (when present) is reset so the closure re-prepares with a fresh pending nonce. Explicit user rejections (EIP-1193 code 4001, matched via viem's typed error classes) are never retried.

### Tests

- `provider.test.ts`: live signer gets a nonce manager + correct address.
- `helpers.test.ts`: retry-on-any-failure with nonce manager reset, retry for signers without a manager, `eth_chainId` fallback when the wallet has no configured chain, user rejection not retried, last error surfaced after retries are exhausted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


alxwlw:
### Note: extra broadcast site beyond the step actions

While reviewing this branch I found that the nonce-retry wrapper needed to cover one more live broadcast that isn't a step action: the on-chain "upgrade-from" info writes.

`writeUpgradeFromInfo` ([`package.ts#L270`](https://github.com/usecannon/cannon/blob/0ff332cd7c929e48bcc212e9b9f8b26db70b2e73/packages/builder/src/package.ts#L270)) runs on every `cannon build ... -- broadcast` and calls `storeWrite` → `deployStoreIfNeeded`, both of which broadcast with the same live signer but were **not** wrapped (unlike `invoke`/`deploy`/`router`/`diamond`/`create2`). The CLI-level retry loop in `build.ts` is not nonce-aware (it never resets the nonce manager), so a stale-nonce failure there would have aborted the build despite the rest of the fix.

Addressed in 0ff332cd7c929e48bcc212e9b9f8b26db70b2e73 — both sends now go through `sendTransactionWithNonceRetry`:

- [`onchain-store.ts#L96-L106`](https://github.com/usecannon/cannon/blob/0ff332cd7c929e48bcc212e9b9f8b26db70b2e73/packages/builder/src/utils/onchain-store.ts#L96-L106) — `deployStoreIfNeeded` (`sendTransaction`)
- [`onchain-store.ts#L111-L126`](https://github.com/usecannon/cannon/blob/0ff332cd7c929e48bcc212e9b9f8b26db70b2e73/packages/builder/src/utils/onchain-store.ts#L111-L126) — `storeWrite` (`writeContract`)

The same commit also tightens the `already known` branch of the nonce matcher to a transaction context so unrelated errors don't trigger a spurious retry.

packages/builder/src/steps/diamond.ts
```diff
@@ -170,7 +171,7 @@ const diamondStep = {
     ]);
 
     // finally, do the cut operation
-    const updateFacets = [];
+    const updateFacets: any[] = [];
```
dbeal-eth: we don't want to expand the valid types to `any` here. can we identify why this would seem to be needed and narrow the type down/preventt this change entirely?

packages/builder/src/helpers.ts
```diff
@@ -77,3 +78,42 @@ export async function getBlockRetried(provider: viem.PublicClient, blockHash: vi
...
+const NONCE_ERROR_RE =
+  /nonce too low|nonce too high|nonce provided.*is (?:too )?(?:low|high|higher|lower)|invalid (?:transaction )?nonce|replacement transaction underpriced|(?:transaction|tx).*already known|already known.*(?:transaction|tx)/i;
+
+// Detects whether an error (or any error in its `cause` chain) is a nonce-related RPC error.
+// viem wraps RPC errors, so the underlying message can live a few levels down in `.cause`.
+export function isNonceError(err: unknown): boolean {
+  let e: any = err;
+  for (let i = 0; e && i < 5; i++) {
+    const msg = e.details ?? e.shortMessage ?? e.message ?? '';
+    if (NONCE_ERROR_RE.test(String(msg))) return true;
```
dbeal-eth: this strategy seems very brittle (either having false positives or negatives). can we instead just *always* retry and assume the nonce is broken? I understand its challenging because there isn't a global or reliable error that RPCs return on this type of error, but we need to have a strategy here that gives maximum predictability to the user. Perhaps we offer a command line flag instead?

packages/builder/src/helpers.ts
```diff
@@ -77,3 +78,42 @@ export async function getBlockRetried(provider: viem.PublicClient, blockHash: vi
...
+  signer: CannonSigner,
+  chainId: number,
+  prepareAndSend: () => Promise<viem.Hash>
+): Promise<viem.Hash> {
+  return (await promiseRetry({ retries: 3, minTimeout: 250, factor: 2 }, (retry) =>
+    prepareAndSend().catch((err: unknown) => {
+      if (!isNonceError(err)) throw err;
+      const account: any = (signer.wallet as any).account;
+      try {
+        account?.nonceManager?.reset?.({ address: account.address, chainId });
```
dbeal-eth: if the account doesn't have its nonce manager set (or it doesn't have a reset function) then this will silently have no effect

packages/builder/src/helpers.ts
```diff
@@ -77,3 +78,42 @@ export async function getBlockRetried(provider: viem.PublicClient, blockHash: vi
...
+// so the happy path is unchanged.
+export async function sendTransactionWithNonceRetry(
+  signer: CannonSigner,
+  chainId: number,
+  prepareAndSend: () => Promise<viem.Hash>
+): Promise<viem.Hash> {
+  return (await promiseRetry({ retries: 3, minTimeout: 250, factor: 2 }, (retry) =>
+    prepareAndSend().catch((err: unknown) => {
+      if (!isNonceError(err)) throw err;
+      const account: any = (signer.wallet as any).account;
```
dbeal-eth: why do we need to do so much type casting here? Seems like an indication that something may be wrong with this strategy.

packages/cli/src/util/provider.ts
```diff
@@ -83,6 +83,22 @@ export const getChainIdFromRpcUrl = async (rpcUrl: string) => {
   return provider.getChainId();
 };
 
+// Builds a live signer from a private key with a viem nonceManager attached, so that nonce
+// assignment is serialized per (address, chainId) and never regresses below an already-used
+// nonce. This prevents "nonce too low" on strict-ordering / eventually-consistent RPCs where a
+// fresh eth_getTransactionCount('pending') read can lag an already-submitted transaction.
+export function createPrivateKeySigner(privateKey: viem.Hex, chainId: number, transport: viem.Transport): CannonSigner {
+  const account = privateKeyToAccount(privateKey, { nonceManager });
```
dbeal-eth: it seems we are using the *default* nonce manager here. Is there any reason why we shouldn't use `createNonceManager()` instead to make sure that the nonces are being managed by the same RPC that is the provided `transport`?

packages/cli/src/util/provider.ts
```diff
@@ -271,16 +279,13 @@ export async function resolveProviderAndSigners({
...
-              wallet: viem.createWalletClient({
-                account,
-                chain: getChainById(chainId),
-                transport: viem.custom(publicClient.transport),
-              }),
-            });
+            signers.push(
+              createPrivateKeySigner(
+                normalizePrivateKey(keyPrompt.value as viem.Hex),
+                chainId!,
```
dbeal-eth: do we know that `chainId` is set here?

packages/website/guides/deploy-onchain.mdx
```diff
@@ -68,6 +68,18 @@ cannon build --rpc-url https://rpc2.sepolia.org --private-key 0xd8...ad
...
+Some RPC providers enforce strict nonce ordering or serve `eth_getTransactionCount` from eventually-consistent / load-balanced replicas. On such endpoints a freshly-read pending nonce can briefly lag an already-submitted transaction. Cannon assigns nonces to live signers through a [viem nonce manager](https://viem.sh/docs/accounts/local/createNonceManager) and retries nonce-class broadcast errors automatically.
+
+If a deploy still aborts partway through, it is safe to re-run it — Cannon resumes where it stopped:
+
+```bash
+cannon build ... --upgrade-from <name:version>
+```
+
+Re-run until the build reports **0 skipped steps**.
+
```
dbeal-eth: this section is probably unneeded, as people expect cannon to be properly managing nonces regardless, and there isn't really anything actionable here by informing users that this feature is in-use.

.changeset/build-broadcast-nonce.md
dbeal-eth, line 1: we have never actually ended up getting around to finishing using changeset in our release process, so you can go ahead and delete this file from the changes.

dbeal-eth (changes requested):
hey! thanks again for the contribution, and sorry for my delay in getting back to you.

There are some things to look at, mostly with regards to type protection and watching the edge cases around null types.

packages/builder/src/steps/diamond.ts
alxwlw: The widening came from moving the send into the retry closure — the array is read inside a nested function, where TS can't apply evolving-array inference. Replaced with an explicit `FacetCut` type and `getFacetSelectors` now returns `viem.Hex[]` via a type-guard filter; `addFacets` in `firstTimeDeploy` is typed the same way. Done in c9bc3fd.

packages/builder/src/helpers.ts
alxwlw: Agreed — done in c9bc3fd. `isNonceError` and the regex are gone; the wrapper (now `sendTransactionWithRetry`) retries every failed broadcast and always assumes the nonce may be broken: before each retry it resets the signer's nonce manager (when present) so the closure re-prepares with a fresh pending nonce. Deterministic failures (e.g. reverts) fail identically on every attempt and surface the original error after the last one, so the only cost there is the bounded backoff (3 retries, ~1.75s total).

The one exception is an explicit user rejection (EIP-1193 code 4001, matched via viem's typed error classes, not strings) — retrying that would re-prompt a wallet user who already declined to sign.

Since unconditional retry behaves predictably without new config surface I didn't add a flag, but happy to add an opt-out if you'd prefer.

packages/builder/src/helpers.ts
alxwlw: Fixed in c9bc3fd — the reset is now an explicit typed path: the account is narrowed to a viem local account and `reset` is only called when a manager is present; both outcomes are debug-logged, so nothing silently no-ops. When there is no manager (json-rpc backed signers, e.g. Frame) there is genuinely nothing to reset client-side — the node assigns their nonces on every send — and the retry still applies.

packages/builder/src/helpers.ts
alxwlw: It was an indication of exactly that — the wrapper was reaching into account internals as `any`. With the account properly narrowed (`account.type === 'local'`), viem fully types `nonceManager` and `reset`, so all casts are gone in c9bc3fd. Dropping the `chainId` parameter (the reset now resolves it the way viem does at consume time: wallet chain, falling back to `eth_chainId`) also removed a bogus `?? 0` fallback in onchain-store.

packages/cli/src/util/provider.ts
alxwlw: Good call. To be precise, neither form owns an RPC: the manager's `jsonRpc` source issues `eth_getTransactionCount` through the client passed at consume time, which is the wallet client built on this `transport` — so even the default manager read through the right RPC. But the module-level singleton is shared process-wide, so 8b15b1a switches to an explicit per-signer `createNonceManager({ source: jsonRpc() })` with a comment documenting that it reads through the same transport the transactions are sent on.

packages/cli/src/util/provider.ts
alxwlw: Yes — within `resolveProviderAndSigners` the `chainId` parameter is a required `number` (it's only optional in the outer `ProviderParams`), so the assertions were redundant; both dropped in 8b15b1a.

packages/website/guides/deploy-onchain.mdx
alxwlw: Removed in 7e803b7.

.changeset/build-broadcast-nonce.md
alxwlw: Deleted in 7e803b7.

packages/builder/src/helpers.ts
alxwlw: Follow-up on the flag question, plus a correction to what I said earlier in this thread.

Retries are now centralised in a single broadcast pipeline (`packages/builder/src/broadcast.ts`) that every live send in the builder goes through — steps, create2, the on-chain upgrade record and `OnChainRegistry`. `wallet.sendTransaction`/`writeContract` appear only in that file, so a new step cannot bypass it, and the duplicated gas-override blocks in four steps collapsed into `ChainBuilderRuntime.sendTransaction`. The policy is user-controlled: `--broadcast-retries <count>` on `cannon build`, default 3, `0` disables. Each retry prints a warning line via a new `Events.BroadcastRetry`.

Two things I got wrong before and want to flag explicitly:

1. Attaching a nonce manager to the signer did nothing on the step paths. viem takes `nonceManager` from the arguments of `prepareTransactionRequest`, not from the account, so `runtime.provider.prepareTransactionRequest(...)` filled the nonce from `eth_getTransactionCount` and the later `wallet.sendTransaction(prepared)` never consulted the manager. The pipeline leaves the nonce out of the prepare, so it is assigned at signing time.

2. The `resetSignerNonce` call I added after your review was a no-op. In viem 2.30 `reset()` clears only `deltaMap`/`promiseMap`, and `get()` already resets those in its own `finally`; the "never regress below the last issued nonce" guarantee lives in `nonceMap`, which `reset()` never touches. It is gone, and no comment claims otherwise.

Because of (2), an unconditional retry with the manager would always sign the next nonce — safe when the earlier attempt reached the mempool, but if it did not (fee below base fee, RPC 5xx after viem's transport retries) it would leave a nonce gap and the skipped transaction would execute on a later build. So retries are now idempotent for private-key signers: the nonce is taken once per logical send, every attempt re-prepares for current gas and fees and re-signs with the same nonce, and before re-sending the node is asked (`getTransaction`) whether an earlier attempt is already known. At most one variant can execute, and no gap is left behind. JSON-RPC signers (Frame) keep the previous behaviour — the node assigns their nonce — so a transient error after acceptance can still prompt a second signature there; `--broadcast-retries 0` opts out.

Tests are in `broadcast.test.ts` against a real viem local account over a custom transport that keeps answering a stale nonce: consecutive nonces across sends while the RPC still says 5, same-nonce re-send after a failure, re-sign with fresh fees at the same nonce, no re-send when the node already knows the transaction, prepare failures retried without consuming a nonce, and the nonce not advanced after retries are exhausted.

One behaviour change worth calling out: `overrides.gasLimit` from cannonfiles was previously forwarded to viem under the key `gasLimit`, which it ignores; it is now passed as `gas`, so a configured limit actually applies on `invoke`/`deploy`/`router` and the diamond create2 path.

PR #1877, opened 2026-06-09
setup deployment with cloudflare wrangler

dbeal-eth:
* installs cloudflare's homegrown "wrangler" tool as a CLI tool in the root
* add a wrangler configuration that explains how to run the commands
* update _redirects to work with cloudflare's expected format.


mjlescano (approved):

packages/website/public/_redirects
```diff
@@ -1,13 +1,14 @@
...
+/packages/:name/:tag/:variant/deployment/contracts /packages/[name]/[tag]/[variant]/deployment/contracts 200
+/packages/:name/:tag/:variant/code /packages/[name]/[tag]/[variant]/code 200
+/packages/:name/:tag/:variant/code/:moduleName /packages/[name]/[tag]/[variant]/code/[moduleName] 200
+/packages/:name/:tag/:variant/interact /packages/[name]/[tag]/[variant]/interact 200
+/packages/:name/:tag/:variant/interact/:moduleName /packages/[name]/[tag]/[variant]/interact/[moduleName] 200
+/packages/:name/:tag/:variant/interact/:moduleName/:contractName/:contractAddress /packages/[name]/[tag]/[variant]/interact/[moduleName]/[contractName]/[contractAddress] 200
+/packages/:name/:tag/:variant/cannonfile /packages/[name]/[tag]/[variant]/cannonfile 200
+/deploy/txn/:chainId/:safeAddress/:nonce/:sigHash /deploy/txn/[chainId]/[safeAddress]/[nonce]/[sigHash] 200
+/address/:chainId/:address /address/[chainId]/[address] 200
+/tx/:chainId/:txHash /tx/[chainId]/[txHash] 200
```
saturn-dbeal, line 14: ⚠️ The `/txs` route was present in the old `_redirects` but is missing here. The page still exists at `packages/website/src/pages/txs/index.tsx`, so without this redirect, `/txs` will 404 on Cloudflare Pages.

Suggest adding:
```
/txs /txs 200
```

wrangler.toml
```diff
@@ -0,0 +1,6 @@
+#:schema node_modules/wrangler/config-schema.json
+name = "cannon"
+compatibility_date = "2026-06-09"
+
+[assets]
+directory = "./packages/website/out"
```
saturn-dbeal, line 6: Consider adding a `[assets.not_found]` configuration for a custom 404 page. Since the website uses Next.js static export with `output: 'export'`, the `404.html` page will be generated in `out/`. Cloudflare Pages can serve this automatically, but you may want to explicitly configure it:

```toml
[assets.not_found]
file = "404.html"
```

This ensures Cloudflare Pages returns the custom 404 page for unmatched routes instead of its default.

package.json
```diff
@@ -45,7 +45,8 @@
     "solhint-plugin-prettier": "^0.0.5",
     "syncpack": "^13.0.4",
     "ts-node": "^10.9.2",
-    "typescript": "^5.8.3"
+    "typescript": "^5.8.3",
```
saturn-dbeal, line 48: Adding `wrangler` to root `devDependencies` is reasonable since it's a deployment tool. The version `^4.98.0` requires Node `>=22.0.0` which aligns with the repo's engine requirement. The lockfile shows it also pulls in `workerd`, `miniflare`, `sharp`, and `esbuild@0.27.3` (upgraded from `0.23.1`) — that esbuild version bump is what causes the large lockfile diff. Worth being aware of in case any contentlayer/webpack builds behave differently.

saturn-dbeal (changes requested):
## Summary

This PR sets up Cloudflare Pages deployment with Wrangler CLI:
- Adds `wrangler` as a root devDependency
- Adds `wrangler.toml` pointing to `packages/website/out`
- Updates `_redirects` to Cloudflare Pages format (removing `.html` extensions)
- Adds new routes for `/learn/guides/get-started/:section` and `/packages/:name/:tag/:variant/deployment/contracts`

## Build & Test Results

✅ **`pnpm install`** — succeeded
✅ **`pnpm build`** — succeeded (core packages compile correctly)
⚠️ **`pnpm test`** — 4 test failures in `packages/cli/src/commands/build.test.ts`, but these are **pre-existing** (verified by running the same tests on `dev` — same failures). Not related to this PR.

## Issues Found

### 🔴 Missing `/txs` redirect (should fix)
The old `_redirects` had `/txs /txs.html 200` but this was dropped in the new version. The page still exists at `packages/website/src/pages/txs/index.tsx`. Without this redirect, `/txs` will return a 404 on Cloudflare Pages.

**Fix:** Add this line to `_redirects`:
```
/txs /txs 200
```

## Notes
- The large `pnpm-lock.yaml` diff (~877 additions) is mostly from wrangler's transitive dependencies, notably an esbuild bump from `0.23.1` → `0.27.3`. This affects contentlayer/webpack resolution paths but builds still pass.
- Base is correctly `dev`.

## Verdict
Looks good overall — just needs the `/txs` redirect added back before merging.

dbeal-eth:
the missing `/txs` redirect is not needed because its redundant-but good catch!

